PR #609, opened 2014-05-19
Adapt riak_test for riak_core applications.

cmeiklejohn:
Adapt riak_test to more easily be used with riak_core based applications.  This involves removing a few assumptions about executable names.


DeadZen:
+1 now #752 needs to be done alongside this

PR #683, opened 2014-09-18
Add more checks on client certificate authentication

macintux:


macintux:
I wrote this early in the effort to investigate POODLE and I simply forgot about it as I got deeper in the weeds. Still needs a review; perhaps @corbaguy could look (and suggest mo' better tests while he's at it).

seancribbs:
@macintux It looks straightforward enough, I'll take it.

tests/pb_cipher_suites.erl
```diff
@@ -182,6 +191,19 @@ confirm() ->
                                                          "127.0.0.1/32",
                                                          "certificate"]]),
 
+    ok = rpc:call(Node, riak_core_console, add_source, [["site4.basho.com",
+                                                         "127.0.0.1/32",
+                                                         "certificate"]]),
+
+    lager:info("Checking legit certificates are allowed"),
+    {ok, PB} = riakc_pb_socket:start("127.0.0.1", Port,
```
seancribbs: This causes a badmatch on line 230 because you don't rename the variable.

corbaguy:
Per our ongoing conversations, I think lots mo' betta' tests would be a Good Thing(TM).

Adding @fadushin to make sure he gets enough spam ;)

PR #692, opened 2014-10-16
Add verify_transfer_limit

andrewjstone:
Gets and sets the transfer limit on a 2 node cluster. Tests both rpc
and cluster metadata based gets.


macintux:
Reviewing

macintux:
This reflects ancient code that has evolved/reworked/dropped. @andrewjstone, is there any reason to keep this open for later reworking?

PR #711, opened 2014-12-12
New cluster/handoff CLI coverage

mrallen1:
Add CLI test coverage for the new handoff and cluster commands added in riak_core from basho/riak_core#659, basho/riak_core#660 and basho/riak_core#661.  These commands include

* handoff enable|disable
* handoff summary
* handoff details
* cluster status
* cluster partitions
* cluster partition-count
* cluster partition id|index
* cluster members

There is also coverage for the "set / show / describe" cuttlefish configuration values in riak_cli.

Also tracking #707 (this PR may conflict depending on merge order.)

### DIFF
--- a/tests/riak_admin_console_tests.erl
+++ b/tests/riak_admin_console_tests.erl
@@ -53,7 +53,20 @@ cluster_tests(Node) ->
     check_admin_cmd(Node, "cluster resize-ring abort"),
     check_admin_cmd(Node, "cluster plan"),
     check_admin_cmd(Node, "cluster commit"),
-    check_admin_cmd(Node, "cluster clear").
+    check_admin_cmd(Node, "cluster clear"),
+    check_admin_cmd(Node, "cluster status"),
+    check_admin_cmd(Node, "cluster partition-count"),
+    check_admin_cmd(Node, "cluster partition-count --node=dev1@127.0.0.1"),
+    check_admin_cmd(Node, "cluster partition-count --node=dev99@127.0.0.1"),
+    check_admin_cmd(Node, "cluster partitions"),
+    check_admin_cmd(Node, "cluster partitions --node=dev1@127.0.0.1"),
+    check_admin_cmd(Node, "cluster partitions --node=dev99@127.0.0.1"),
+    check_admin_cmd(Node, "cluster partition id=0"),
+    check_admin_cmd(Node, "cluster partition index=0"),
+    check_admin_cmd(Node, "cluster partition id=-1"),
+    check_admin_cmd(Node, "cluster partition index=-1"),
+    check_admin_cmd(Node, "cluster members"),
+    check_admin_cmd(Node, "cluster members --all").
 
 %% riak-admin bucket_type
 bucket_tests(Node) ->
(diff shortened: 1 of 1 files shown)


macintux:
Reviewing

macintux:
Looks like something has changed. Here's a test run against EE 2.0.4 tag:

```
09:54:57.616 [info] "---- Cluster Status ----\nRing ready: true\n\n+--------------------+------+-------+-----+-------+\n|        node        |status| avail |ring |pending|\n+--------------------+------+-------+-----+-------+\n| (C) dev1@127.0.0.1 |valid |  up   |100.0|  --   |\n+--------------------+------+-------+-----+-------+\n\nKey: (C) = Claimant; availability marked with '!' is unexpected\n"
09:54:57.616 [error] Error in process <0.189.0> on node 'riak_test@127.0.0.1' with exit value: {{assertEqual_failed,[{module,riak_admin_console_tests},{line,283},{expression,"Out"},{expected,"pass"},{value,"---- Cluster Status ----\nRing ready: true\n\n+--------------------+------+-------+-----+-------+\n|   ...


09:54:57.616 [warning] riak_admin_console_tests failed: {{assertEqual_failed,[{module,riak_admin_console_tests},{line,283},{expression,"Out"},{expected,"pass"},{value,"---- Cluster Status ----\nRing ready: true\n\n+--------------------+------+-------+-----+-------+\n|        node        |status| avail |ring |pending|\n+--------------------+------+-------+-----+-------+\n| (C) dev1@127.0.0.1 |valid |  up   |100.0|  --   |\n+--------------------+------+-------+-----+-------+\n\nKey: (C) = Claimant; availability marked with '!' is unexpected\n"}]},[{riak_admin_console_tests,'-check_admin_cmd/2-fun-1-',2,[{file,"tests/riak_admin_console_tests.erl"},{line,283}]},{riak_admin_console_tests,cluster_tests,1,[{file,"tests/riak_admin_console_tests.erl"},{line,57}]},{riak_admin_console_tests,confirm,0,[{file,"tests/riak_admin_console_tests.erl"},{line,272}]},{riak_test_runner,return_to_exit,3,[{file,"src/riak_test_runner.erl"},{line,159}]}]}
09:54:57.617 [error]
================ riak_admin_console_tests failure stack trace =====================
{{assertEqual_failed,[{module,riak_admin_console_tests},
                      {line,283},
                      {expression,"Out"},
                      {expected,"pass"},
                      {value,"---- Cluster Status ----\nRing ready: true\n\n+--------------------+------+-------+-----+-------+\n|        node        |status| avail |ring |pending|\n+--------------------+------+-------+-----+-------+\n| (C) dev1@127.0.0.1 |valid |  up   |100.0|  --   |\n+--------------------+------+-------+-----+-------+\n\nKey: (C) = Claimant; availability marked with '!' is unexpected\n"}]},
 [{riak_admin_console_tests,'-check_admin_cmd/2-fun-1-',2,
                            [{file,"tests/riak_admin_console_tests.erl"},
                             {line,283}]},
  {riak_admin_console_tests,cluster_tests,1,
                            [{file,"tests/riak_admin_console_tests.erl"},
                             {line,57}]},
  {riak_admin_console_tests,confirm,0,
                            [{file,"tests/riak_admin_console_tests.erl"},
                             {line,272}]},
===================================================================================
```

mrallen1:
Hmm, this looks a bit hairy. Will have a think on it.

andrewjstone:
@mrallen1 I think I see your confusion. I took a look at the tests as well and the output of those function calls can't possibly match `Val`. I then realized that `Val` is the list of arguments passed *to* the function being intercepted. Therefore these tests are only testing that the commands parsed by the shell result in the right callback function firing when called from bash via nodetool.

In our case, since we only call one function and the parsing and callbacks are handled by clique, we'd have to make sure we intercept the registered callback called from clique. However, since clique handles parse errors automatically and won't call the callback in that case, we couldn't do negative testing this way. It would however test that the commands are actually registered correctly and get called when given a correct command line string. I'd give that a try.

andrewjstone:
I'm now wondering what happens if the intercept doesn't fire because a bad string is passed. I was thinkng there would be a timeout, and the test would fail, but now I'm not so sure. I bet if it's just not called and the test passes. In that case it makes these tests pretty useless, since you can't tell when they actually pass or fail. Something to look into @mrallen1.

mrallen1:
OK, thanks @andrewjstone - I will check into it more closely tomorrow.

andrewjstone:
@mrallen1 After looking at the comment at the top of this file, which I should have done in the first place, I realize that this test is only intended to check for bash portability bugs via intercepts. That's why it only checks the input to the intercepted function, as I noted in my previous comments.

Your tests check the output of the calls to ensure they return something. So while they do indeed prove that a correct round trip call occurred, they are quite brittle as they break with a changing message, which is not really the thing we are gaining by testing with them.

As noted, handoff commands aren't directly parsed by bash, so the original purpose of the test is not necessary for our new clique based commands. Since they aren't really useful for the purpose of this riak_test, and are very brittle, I'd say we just don't bother updating this test the new cli stuff at all.

macintux:
Mark and I discussed the incongruity of his tests vs what was being done already; thanks for catching why.

I agree, makes more sense rely on clique's tests. Fortunately, Mark and I both learned a fair bit about the intercepts and riak_test in general while working on this, so definitely not wasted energy.

mrallen1:
OK, in light of that, I'm going to close this PR. Thanks for the feedback.

PR #713, opened 2014-12-16
check riak-admin status

uwiger:
Update verify_riak_stats to also compare the outputs of HTTP stats and `riak-admin status`

A workaround implemented to compensate for the rt:admin/2 function dropping parts of large output.


jburwell:
@seancribbs yes.  This test confirms that all of the stats emitted by the HTTP endpoint are present in `riak-admin status`.  The fact the two have been out-of-sync has been a long standing nit by a number of folks.

jburwell:
I have run the test locally, and it passes.

PR #722, opened 2014-12-30
Added test for realtime helper process leak.

lordnull:
Test uses the method used in the report of the issue at https://github.com/basho/riak_repl/issues/649. It is simply r_t'ified.


jonmeredith:
+1 7539ad5

verifies fix, would be better extending the EQC test to find more general cases.

Merging by hand as no bors on riak_test.

PR #725, opened 2015-01-09
Fix several framework bugs found while testing

kellymclaughlin:
* Address two issues related to bucket type creation in rt_cluster
* Address an issue with rt:stream_cmd_loop using an outdated rt:config
call
* Adjust the expected arity of the setup function in riak_test_runner
from 2 to 1
* Change rtdev:get_deps/0 to use the default_version from the config
instead of a hardcoded path


javajolt:
:+1:

PR #726, opened 2015-01-09
Add an external_properties field to rt_properties record

kellymclaughlin:
Add an external_properties field to rt_properties record to facilitate
testing with applications built on Riak. This field can be use for
arbitrary data by external applications and could be useful for
testing things like Riak CS.


javajolt:
:+1:

PR #728, opened 2015-01-21
Test for basho/riak#667, read pre 2.0.4 sets/maps

russelldb:
Requires https://github.com/basho/riak_dt/pull/111

### DIFF
--- a/tests/kv667.erl
+++ b/tests/kv667.erl
@@ -1,0 +1,103 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2012 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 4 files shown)


tests/kv667.erl
```diff
@@ -0,0 +1,122 @@
...
+    Map3 = riakc_map:update(
+            {<<"names">>, set},
+            fun(R) ->
+                    riakc_set:add_element(<<"Updated">>, R)
+                    end, O),
+
+    Map4 = riakc_map:update({<<"profile">>, map},
+                            fun(M) ->
+                                    riakc_map:update(
+                                      {"name", register},
```
andrewjstone: I thought the key had to be a binary here, but this appears to work.

andrewjstone:
+1
The code looks good. While using a previous version of 2.0.2, the test fails with 2.0 branch of riak_dt and passes with basho/riak_dt#111

PR #729, opened 2015-01-21
Updating rebar.config to be set purely to tags/SHAs

gcymbalski:
As requested following conversations with Andrew Stone, here's a PR to pin deps to the latest tags available in subrepos. Be aware that work may be needed in those subrepos to actually bring in commits that are expected, as previous builds have been against branches.


Basho-JIRA:
See comments here for additional detail - https://github.com/basho/riak/pull/669

These issues should just be closed without merging and we should just use the current release strategy (for now).


Closing the bug as we will not include the updates. 

***[posted via JIRA by Derek Somogyi]***

PR #730, opened 2015-01-22
Update cs test runner for Riak CS 2.0

ksauzz:


src/rt_cs_dev.erl
```diff
@@ -78,7 +78,7 @@ upgrade(Node, NewVersion) ->
     N = node_id(Node),
     Version = node_version(N),
     lager:info("Upgrading ~p : ~p -> ~p", [Node, Version, NewVersion]),
-    stop(Node),
+    catch stop(Node),
```
kuenishi: What is this catch for?

ksauzz: The catch is for the error when riak already stooped.

kuenishi:
+1

PR #734, opened 2015-02-11
Refactor tests/overload to use less magical-number comparisons

aberghage:
Since this test often fails due to fuzzy magic number comparisons (e.g. 1.2*?THRESHOLD) when trying to determine whether or not overload thresholds are being correctly enforced. Since these aren't exact, it seems that a decent approach is to, for now, ask "did any work get shed" instead — this changeset does that (and a little refactoring to reduce repetition).

In the future we should revisit this and just explicitly track denials instead of fsm processes / vnode queue depth, as @slfritchie initially suggested.

### DIFF
--- a/tests/overload.erl
+++ b/tests/overload.erl
@@ -103,7 +103,7 @@ test_vnode_protection(Nodes, BKV, ConsistentType) ->
     rt:pmap(fun(Node) ->
                     rt:update_app_config(Node, Config)
             end, Nodes),
-    ProcFun = build_predicate_lt(test_vnode_protection, (?NUM_REQUESTS), "ProcFun", "Procs"),
+    ProcFun = build_predicate_lt(test_vnode_protection, (2*?THRESHOLD * 1.5), "ProcFun", "Procs"),
     QueueFun = build_predicate_lt(test_vnode_protection, (?NUM_REQUESTS), "QueueFun", "QueueSize"),
     verify_test_results(run_test(Nodes, BKV), ConsistentType, ProcFun, QueueFun),
 
--- a/tests/overload.erl
+++ b/tests/overload.erl
@@ -77,14 +77,10 @@ test_no_overload_protection(_Nodes, _BKV, true) ->
     ok;
 test_no_overload_protection(Nodes, BKV, ConsistentType) ->
     lager:info("Testing with no overload protection"),
-    ProcFun = fun(X) ->
-                      lager:info("in test_no_overload_protection ProcFun, Procs:~p, Metric:~p", [X, ?NUM_REQUESTS]),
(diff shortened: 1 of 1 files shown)


tests/overload.erl
```diff
@@ -107,14 +103,8 @@ test_vnode_protection(Nodes, BKV, ConsistentType) ->
...
             end, Nodes),
-    ProcFun = fun(X) ->
-                      lager:info("in test_vnode_protection ProcFun, Procs:~p, Metric:~p", [X, (2*?NUM_REQUESTS * 1.5)]),
-                      X =< (2*?THRESHOLD * 1.5)
-              end,
-    QueueFun = fun(X) ->
-                      lager:info("in test_vnode_protection QueueFun, QueueSize:~p, Metric:~p", [X, (?NUM_REQUESTS * 1.1)]),
-                      X =< (?THRESHOLD * 1.1)
-              end,
+    ProcFun = build_predicate_lt(test_vnode_protection, (2*?THRESHOLD * 1.5), "ProcFun", "Procs"),
```
andrewjstone: I don't really understand why # of procs wouldn't always be equal to ?NUM_REQUESTS. I probably just don't understand how overload_proxy actually fits in here.

andrewjstone: Oh, I see it's used via the intercept.

andrewjstone: Shouldn't it always be equal to Threshold then?

tests/overload.erl
```diff
@@ -127,15 +117,10 @@ test_vnode_protection(Nodes, BKV, ConsistentType) ->
...
-    ProcFun2 = fun(X) ->
-                      lager:info("in test_vnode_protection after suspend ProcFun, Procs:~p, Metric:~p", [X, ?NUM_REQUESTS]),
-                      X >= ?NUM_REQUESTS
-
-              end,
-    QueueFun2 = fun(X) ->
-                      lager:info("in test_vnode_protection after suspend QueueFun, QueueSize:~p, Metric:~p", [X, (?THRESHOLD * 1.1)]),
-                      X =< (?THRESHOLD * 1.1)
-              end,
+    ProcFun2 = build_predicate_gte("test_vnode_protection after suspend",
```
andrewjstone: How can we have more GET fsm processes than requests?

aberghage: As far as I know we can't, but since it's gte it's basically just defensive coding. Or at least that's how I read the original, which made enough sense to keep I think

tests/overload.erl
aberghage: Pretty sure you're right — I'll double-check and change it. No sense in propagating magic numbers along if we know what they actually should be.

andrewjstone:
+1

PR #735, opened 2015-02-11
Ensure that setting advanced config on broken path does not cause an err...

andrewjstone:
...or


cmeiklejohn:
Reviewing now...

cmeiklejohn:
:+1:

PR #736, opened 2015-02-11
Change verify_dt_converge to use riakc_flag:enable(F) rather than disable.

JeetKunDoug:
This will work even in the rare and pathological case where the cluster is partitioned before
all 3 nodes have received the update. riakc_flag:disable(F) requires context, which isn’t there in the 
new map that would be created on the side of the partition with no data.


aberghage:
+1

aberghage:
make that +1 04d54cb

PR #737, opened 2015-02-11
Better support for wildcard certificates:

JeetKunDoug:
- Update make_certs to support wildcard certificates by passing an additional Path parameter.
- add replication2_ssl test to ensure using wildcard certificates works after PR to riak_core_ssl_util.

### DIFF
--- a/src/make_certs.erl
+++ b/src/make_certs.erl
@@ -112,8 +112,14 @@ intermediateCA(Root, CA, ParentCA) ->
 endusers(Root, CA, Users) ->
     [enduser(Root, CA, User) || User <- Users].
 
-enduser(Root, CA, User) -> 
-    UsrRoot = filename:join([Root, User]),
+enduser(Root, CA, {User, Path}) ->
+    enduser(Root, CA, User, Path);
+
+enduser(Root, CA, User) ->
+    enduser(Root, CA, User, User).
+
+enduser(Root, CA, User, Path) ->
+    UsrRoot = filename:join([Root, Path]),
     file:make_dir(UsrRoot),
     CnfFile = filename:join([UsrRoot, "req.cnf"]),
     DN = #dn{commonName = User},
--- a/src/make_certs.erl
+++ b/src/make_certs.erl
@@ -112,14 +112,14 @@ intermediateCA(Root, CA, ParentCA) ->
 endusers(Root, CA, Users) ->
     [enduser(Root, CA, User) || User <- Users].
 
-enduser(Root, CA, {User, Path}) ->
-    enduser(Root, CA, User, Path);
+enduser(Root, CA, {User, DirName}) ->
+    enduser(Root, CA, User, DirName);
 
 enduser(Root, CA, User) ->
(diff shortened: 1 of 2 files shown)


JeetKunDoug:
Depends on https://github.com/basho/riak_core/pull/701, which is being held until 2.0.5 is released, so please DON'T MERGE yet.

src/make_certs.erl
```diff
@@ -112,8 +112,14 @@ intermediateCA(Root, CA, ParentCA) ->
 endusers(Root, CA, Users) ->
     [enduser(Root, CA, User) || User <- Users].
 
-enduser(Root, CA, User) -> 
-    UsrRoot = filename:join([Root, User]),
+enduser(Root, CA, {User, Path}) ->
```
tburghart: Minor, but I'd prefer if the `Path` parameter was named something like `DirName`, as it's only one segment of the filesystem path.

src/make_certs.erl
```diff
@@ -112,8 +112,14 @@ intermediateCA(Root, CA, ParentCA) ->
...
+    enduser(Root, CA, User, Path);
+
+enduser(Root, CA, User) ->
+    enduser(Root, CA, User, User).
+
+enduser(Root, CA, User, Path) ->
```
tburghart: Same comment re: name of `Path` parameter.

tests/replication2_ssl.erl
```diff
@@ -188,6 +189,25 @@ confirm() ->
...
+    SSLConfig8 = [
+        {riak_repl,
+            [
+                {fullsync_on_connect, false},
+                {fullsync_interval, disabled}
+            ]},
+        {riak_core,
+            [
+                {ssl_enabled, true},
+                {peer_common_name_acl, ["*.basho.com"]},
```
tburghart: Simple wildcard test should not use an ACL.

tests/replication2_ssl.erl
```diff
@@ -253,6 +273,12 @@ confirm() ->
     ?assertEqual(ok, test_connection({Node1, merge_config(SSLConfig5, BaseConf)},
             {Node2, merge_config(SSLConfig6, BaseConf)})),
 
+
+    lager:info("===testing using same wildcard certificate"),
+    rt:log_to_nodes([Node1, Node2], "wildcard certificates test"),
+    ?assertEqual(ok, test_connection({Node1, merge_config(SSLConfig8, BaseConf)},
+        {Node2, merge_config(SSLConfig8, BaseConf)})),
+
```
tburghart: Should have a test connecting wildcarded cert to non-wildcarded cert, without ACL, such as `SSLConfig3` <==> `SSLConfig8`

tests/replication2_ssl.erl
JeetKunDoug: Not sure why GitHub left this as a "valid" comment - SSLConfig8 now has no ACL, and I added 9 to make sure we tested wildcard certs with ACLs

tburghart:
+1 137a11c

PR #741, opened 2015-02-26
Assert build_type to prevent setup mistakes

shino:
Riak CS uses OS environment variable to choose build type, oss or ee.
On the other hand, riak_test config includes `build_type` to express
test target. This commit adds assertion that cs_current and cs_previous
is actually build as `build_type` indicates. It also excludes oss/ee discrepancy
between cs_current and cs_previous.

RCS-112


kuenishi:
+1, or is @borshop enabled here?

shino:
Merged manually

PR #742, opened 2015-03-03
Ensure cover:start is called after verifying node is pingable

darach:
The cover server was failing to start until I added this wait in rt_cover in a
fork of riak_test for a riak_core based app.


cmeiklejohn:
:+1:

PR #744, opened 2015-03-07
Add test that always passes. [JIRA: RIAK-1588]

tburghart:
When included in an automatically generated list of tests that may only contain one test, ensures riak_test will continue if that test fails.

### DIFF
--- a/tests/always_pass_test.erl
+++ b/tests/always_pass_test.erl
@@ -1,0 +1,8 @@
+%% @doc A test that always returns `pass'.
+-module(always_pass_test).
+-behavior(riak_test).
+-export([confirm/0]).
+
+-spec confirm() -> pass | fail.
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
Can you clarify further why this is needed?

tburghart:
@cmeiklejohn I have scripts that collect tests to run based on assorted rules, and some groupings can result in exactly one test being run.  If that test fails, riak_test blocks waiting for an answer to whether to leave the node(s) running.  Adding a test that is guaranteed to pass bypasses the explicit `{1, fail}` case that causes r_t to block, so when generating a command line, if the number of tests is exactly one then I have a means to make sure r_t will return to my script in finite time.

cmeiklejohn:
:+1:  4527474

PR #746, opened 2015-03-11
Add xmerl_version to the list of expected stats

nickelization:
This fixes some failures we were seeing with the verify_riak_stats test.


javajolt:
:+1:  e2c8ec7

PR #749, opened 2015-03-12
Remove code:which/1 from pb_security and http_security

javajolt:
Currently `code:which/1` relies on specific settings in the Erlang path.  Replacing it with `code:get_object_code/1` seems to be more flexible and will work better with the up and coming testing framework.

**NOTE:** `http_security` seems to be rather fragile and does not pass every time.  YMMV.


seancribbs:
+1

Issue #260, opened 2013-04-06
generalizing riak_test [JIRA: RIAK-2092]


Licenser:
Hi,
I have played a bit with riak_test and noticed it's very tightly tied to riak_kv, which I personally think is a bit of a shame. The framework itself is awesome and I think could be used far beyond the realm of riak_kv, it's qutie easy to adapt for riak_core based applications but I think it would be posible to generalize/moduralize it to a point where it would be useful for testing any kind of distributed erlang application.

That said it's not a blind 'please do this for me NOW!' feature request I'm planning to invest effort into this given there is a general interest into this and a decent chance to get a setup like this merged.

What I was thinking about is making riak_test somewhat of a library with 'extending' libraries that specialise on certain aspects.

A rough idea would be something like this:
- riak_test
  - provides the base infrastructure of running tests, logging validating
  - provides methods to run shell commands, remote commands on nodes, spawn nodes etc.
- riak_core_test
  - requires riak_test
  - provides some general tests that should work on all riak_core applications
  - provide functions to manipulate riak_core applications (build the ring, join nodes, take them down etc)
- riak_kv_test
  - requires riak_core_test
  - provides riak_kv tests
  - provides functions specific to riak_kv

It would then be nice to have something like a riak_test rebar plugin which sets up a basik testing framework in which a user then can just require riak_core_tests or riak_kv_tests and build additional settings based on that. It would allos allow later on to make some more libraires that could dynamically be included that abstract over other aspects of testing.

Cheers,
Heinz

darach:
+1 I've just started following Heinz's recipe as we're adopting riak_core for a project. Contributing fixes/patches would be easier if the core framework was more easily reusable.

cmeiklejohn:
We've adapted riak_core for both SyncFree's antidote and lasp on the master branch.  I'm curious what work is not finished that would need to be.

darach:
Good question. A little documentation would go a long way. Heinz's blog article is useful, for example. I have fairly vanilla riak_test working with adding a custom harness. Some of the tests would be nice to reuse. For example, the ensemble tests depend on riak_kv and a riak_kv specific ensemble backend. A subset that targeted riak_ensemble over riak_core with the default backend would be useful too. 

[...]

PR #752, opened 2015-03-20
Make default riak services and console configurable in rt.erl

darach:
As mentioned in #260 I had another change to riak_test I needed to perform, following
@Licenser's blog article to adapt it to my current riak_core based project. This PR reflects
that change. In summary it

* Makes the riak services that riak_test ensures are started configurable. This defaults to [riak_kv]
* Makes the riak console module configurable. This defaults to riak_kv_console.
* Adds an rt_services configuration variable
* Adds an rt_console configuration variable

The defaults remain unchanged so there should be no breaking introduced.

With this change it should be possible to precompile a self contained riak_test escript and
include it in a project in much the same way as rebar or relx. In this way the riak_core based
project can provide its own rt_harness and configuration and will have a reduced need to use
a forked version of riak_test. @loucash is already using riak_test a compiled escript and this
works well for his project.

A PR based on or derived from this PR would allow us to whack our riak_test forks and use
basho's riak_test vanilla out of the box without changes.

FYI @cmeiklejohn - The only different then to the lasp usage of riak_test would be using a self-contained riak_test escript to drive the tests rather than a local installation. So we may even be able to run riak_test under travis!

### DIFF
--- a/examples/riak_test.config
+++ b/examples/riak_test.config
@@ -46,6 +46,17 @@
     %% scripts in the bin/ directory.
     {rt_harness, rtdev},
 
+    %% The riak test environment can ensure that a default set of riak
+    %% services are started. The set of services can be configured by
+    %% configuring the desired set of services. To alter the default
+    %% enable and configure the following variable
+    %% {rt_services, [riak_kv]},
+
+    %% The riak test environment integreates with riak_kv_console by
+    %% default. To override this for other riak_core based apps uncomment
+    %% and configure teh following variable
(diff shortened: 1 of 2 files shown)


loucash:
This is very useful and a good step forward to separate riak_test from riak_kv.

DeadZen:
+1 This definitely needs to go in immediately

mjbrender:
For instance @seemaj, this PR is a great fix that needs a few minutes of attention. Thanks for helping me prioritize this work!

PR #753, opened 2015-03-23
Use rt:waitfor on reads in ensemble-byzantine test

nickelization:
There were some race conditions we found in this test where reads to an ensemble may fail for a brief period of time after recovering from certain byzantine failures. By using the waitfor function, we ensure the tests will still pass as long as things eventually recover and start working again.


seancribbs:
+1

PR #754, opened 2015-03-26
Set proper exit status

loucash:
I am using `riak_test` on Travis CI and properly set exit status would be very useful.

### DIFF
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -181,7 +181,7 @@ main(Args) ->
 
     Teardown = not proplists:get_value(keep, ParsedArgs, false),
     maybe_teardown(Teardown, TestResults, Coverage, Verbose),
-    ok.
+    proper_exit_status(TestResults).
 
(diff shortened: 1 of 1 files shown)


mjbrender:
This is fantastic @loucash! Thank you for it. Let me check w/the team before making any merges - we're a little behind on PRs for `riak_test` and appreciate your patience.

loucash:
Cool, thanks :+1:

src/riak_test_escript.erl
```diff
@@ -373,6 +373,15 @@ print_summary(TestResults, CoverResult, Verbose) ->
     end,
     ok.
 
+proper_exit_status(TestResults) ->
+    FailCount = length(lists:filter(fun(X) -> proplists:get_value(status, X) =:= fail end, TestResults)),
```
seancribbs: A list comprehension would be clearer here, and you also don't need to compute the length in the following case expression if you just match the empty list.

seancribbs:
:+1:

PR #757, opened 2015-04-01
DP-6 add api_entrypoints test

hmmr:


mdigan:
Closing this pull request. We're taking a different approach. John and Mark will be enhancing the coverage plan API, and we'll be adding functionality in the java client to use the new API features, and using all of those in the spark connector.

PR #758, opened 2015-04-01
Feature/fd/riak 1534

fadushin:
This branch contains two new riak tests:

verify_write_once
verify_handoff_write_once

These tests exercise the put/get path for "write_once" buckets.


seancribbs:
+1 merging. We need these to do testing.

PR #759, opened 2015-04-02
Fix Pid Registration of riak_repl2_leader_intercepts:set_leader_node/1

javajolt:
If this interceptor is used in more than one test in a test suite it fails because it tries to reregister an existing Pid.  Now first always unregister before registering.


seancribbs:
:+1:

PR #760, opened 2015-04-06
Added explicit wait_for_service riak_pipe on pipe* tests

aberghage:
Some of the pipe tests use the arity-1 version of `rt:deploy_nodes`, which doesn't implicitly wait for any Riak services (so it's possible for them to race on pipe being available). This patch adds an explicit wait for all nodes to those tests.

### DIFF
--- a/tests/pipe_verify_basics.erl
+++ b/tests/pipe_verify_basics.erl
@@ -42,6 +42,8 @@ confirm() ->
     lager:info("Build ~b node cluster", [?NODE_COUNT]),
     Nodes = rt:build_cluster(?NODE_COUNT),
 
+    [rt:wait_for_service(Node, riak_pipe) || Node <- Nodes],
+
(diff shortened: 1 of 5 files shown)


seancribbs:
Is it possible for these to use the version of `build_cluster` or `deploy_nodes` that specifically waits for `riak_pipe`?

aberghage:
It doesn't look like [`build_cluster`](https://github.com/basho/riak_test/blob/master/src/rt.erl#L1053-L1077) has an arity that allows specifying a service list, unfortunately. I can add that if you'd prefer that method of fixing this. It is somewhat non-obvious to new test writers that they should need to wait on particular services by hand...

aberghage:
(or I can rework the initialization in these tests to use deploy_nodes instead)

seancribbs:
No problem, then. I'm ok with the change as-is.

seancribbs:
:+1:

PR #761, opened 2015-04-06
Fix BTA-174

seancribbs:
Failure seen at
http://giddyup.basho.com/#/projects/riak_ee/scorecards/120/120-157-verify_riak_stats-centos-6-64/89645

* Typo in one stat name was corrected since
  2.0.5 (search_index_throughtput_one -> search_index_throughput_one)
* write_once code path added its own stats.
* Adjusted assertion at comparison point so that the differences appear
  in the assertion failure as well as the logs.
* Sorted the entries in the list of stats.
* Cleaned up trailing whitespace.


fadushin:
+1 40a583a798613134b612fad0c23e685b4a7bf2ca

PR #762, opened 2015-04-06
updates to handoff for shell script path and list correct var

zeeshanlakhani:
tests need to be green.

### DIFF
--- a/tests/yz_handoff.erl
+++ b/tests/yz_handoff.erl
@@ -125,7 +125,9 @@ confirm() ->
 %%%===================================================================
 
 spawn_handoff_executable(Env) ->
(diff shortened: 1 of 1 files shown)


tests/yz_handoff.erl
```diff
@@ -125,7 +125,9 @@ confirm() ->
 %%%===================================================================
 
 spawn_handoff_executable(Env) ->
-    erlang:open_port({spawn_executable, "handoff-test.sh"},
+    {ok, CurrDir} = file:get_cwd(),
```
seancribbs: How do you know that it's in the current directory?

seancribbs:
Discussed with @zeeshanlakhani, it would be better to move the script into the test itself so that it's reproducible without relying on paths.

PR #763, opened 2015-04-06
Force use of UTF8 for cert fields to avoid OpenSSL incompatibility issues.

JeetKunDoug:
Update and uncomment string_mask to ensure we create supported certificate fields. Without, certain automated builders will create subject CN fields that are of type “teletexString” which has been deprecated since 2003.


seancribbs:
+1 

This is BTA-164.

PR #764, opened 2015-04-06
Add string_mask=utf8only to CA as well, since "basho" and "basho" don't match

JeetKunDoug:
when one is UTF8 and the other is still teletex.


seancribbs:
+1

seancribbs:
This is BTA-164

PR #765, opened 2015-04-06
Fix BTA-172.

seancribbs:
Previous to 2.0, the CRDT capability was exactly `[pncounter]`. However, when upgrading from 2.0 to 2.1, the capability *includes* `pncounter` but is not exactly equal, resulting in a hang in `populate_counters/1`. We just want the test to be sure that it is capable of `pncounter`, not exactly equal, which will make the test continue on 2.0+ clusters.


JeetKunDoug:
:+1:

PR #766, opened 2015-04-06
Revert "Set proper exit status"

seancribbs:
Reverts basho/riak_test#754

The new world being created by @jburwell @javajolt uses specific exit codes for specific failure types. Reverting so as not to screw up their work. Sorry @loucash.

### DIFF
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -181,7 +181,7 @@ main(Args) ->
 
     Teardown = not proplists:get_value(keep, ParsedArgs, false),
     maybe_teardown(Teardown, TestResults, Coverage, Verbose),
-    proper_exit_status(TestResults).
+    ok.
 
 maybe_teardown(false, TestResults, Coverage, Verbose) ->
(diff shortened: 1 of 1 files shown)


loucash:
Ok, no worries.

jburwell:
@loucash @seancribbs the next revision of r_t will support reporting skipped and failed tests where by a skipped test is a test case that can not be run due to environmental incompatibilities (e.g. the requires a backend that is not being tested, the test case does not support the version being tested, etc).  Therefore, a return code of 1 will indicate that one or more test cases were skipped, and 2 will indicate that one or more tests failed.  Since we are trying to maintain as much backwards compatibility as possible, we didn't want to change the meaning of a return of 1.

I will look into modifying this patch to be compatible with the upcoming return code scheme in order to smooth the way for the next release, and also, give you need a much needed capability.

loucash:
@jburwell Cool, yeah, I don't mind return code 2 for tests failed :+1: Cheers!

PR #767, opened 2015-04-06
refactor yz_handoff test to use only rtisms

zeeshanlakhani:
* and remove need for shell script and path needs. 

Fixes BTA-176.


seancribbs:
Code seems fine, but as discussed in chat we need to lower the `vnode_management_timer` because on multiple runs it was possible to get into a wedged state that exceeded the default wait_until timeout.

seancribbs:
This seems like the best we can do for now. :+1:

PR #768, opened 2015-04-07
Make verify_membackend work with Riak 2.1

macintux:
Two key changes:
- The path into the `kv_vnode` state under 2.1
- The target memory values are off by 4 bytes

Going to look into the latter, but the values are predictable through the test run.


macintux:
Not working with a local 2.0.4 instance, so digging deeper

mrallen1:
@javajolt has a better update in #821

PR #769, opened 2015-04-08
Added case clause for case in which we are calling get on a 2.0 or later node

fadushin:
The previous behavior was targeted for pre-2.0 (pre-bucket type) behavior.


JeetKunDoug:
:+1:

PR #770, opened 2015-04-08
Added test for AAE full sync and bucket types

bsparrow435:
This change adds a test for aae fullsync strategy after it was
discovered to not work with bucket types in 2.0+.

### DIFF
--- a/tests/repl_bucket_types.erl
+++ b/tests/repl_bucket_types.erl
@@ -39,7 +39,13 @@ setup(Type) ->
             rt:wait_until_bucket_type_status(DefinedType, active, BNodes),
             rt:wait_until_bucket_type_visible(BNodes, DefinedType);
         mixed ->
-            ok
+            ok;
+        aae ->
+            rt:create_and_activate_bucket_type(LeaderB,
+                                               DefinedType,
+                                               [{n_val, 3}, {allow_mult, false}]),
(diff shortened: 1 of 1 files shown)


bsparrow435:
665 was merged so this guy should be good to go against develop. It will fail for anything using develop of riak_repl though.

cmeiklejohn:
@bsparrow435 

What should this test look like when it fails?  I want to make sure I'm observing the right failure condition.

bsparrow435:
It should hang until the riak_test dies. The AAE fullsync will retry infinitely with bucket types.

cmeiklejohn:
@bsparrow435 

OK, seems good. 

However, we need to split this out into a separate test, given it has to be versioned with GiddyUp so it only runs on releases after 2.1. 

Can you take care of this?

bsparrow435:
Ahhhh, that's a very good point hadn't thought about that. Yes, I can split this out into a specific test. Will get to it tonight.

cmeiklejohn:
OK, cool.

Let me know once the PR is open for the change and I'll do a review there as well.

bsparrow435:
I'll keep it on this branch/PR, I'll just revert the changes to the existing test and make a new one. Will ping you when it's done.

cmeiklejohn:
Obsoleted by https://github.com/basho/riak_test/pull/797.

PR #771, opened 2015-04-08
Removing obsolete repl_reduced test.

JeetKunDoug:
The feature for which it was written will never be shipped. Fixes BTA-167 and perhaps other intermittent failures, as this test would continue running after riak_test thought it had killed it.


macintux:
Also need a database update, @JeetKunDoug is looking at that.

:+1:

JeetKunDoug:
See https://github.com/basho/giddyup/pull/150 for related SQL update for giddyup.

fadushin:
+1 ad91ce48b41c0b213db7c7963b306ea148f05702

javajolt:
+1000 (for good measure)

PR #772, opened 2015-04-08
Bugfix/BTA-183

fadushin:
Fixed two capability tests (verify_counter_capabilty and verify_crd_capability) to work with previous, 2.0.5 and legacy 1.4.12.  Both tests had assumptions about capabilities in previous point releases, which are no longer valid assumptions.  Also relaxed the requirement for capability checking by checking for subsets instead of exact matches (in case new capabilities are added in the future).

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -853,21 +853,34 @@ capability(Node, Capability, Default) ->
 wait_until_capability(Node, Capability, Value) ->
     rt:wait_until(Node,
                   fun(_) ->
-                          cap_equal(Value, capability(Node, Capability))
+                      Cap = capability(Node, Capability),
+                      lager:info("Capability on node ~p is ~p~n",[Node, Cap]),
+                      cap_equal(Value, Cap)
                   end).
 
 wait_until_capability(Node, Capability, Value, Default) ->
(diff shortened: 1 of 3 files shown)


javajolt:
@fadushin Will it still work with legacy=1.3.4 and previous=1.4.12?

JeetKunDoug:
:+1:

fadushin:
Brett, no they won't.  The tests were implemented to be version-dependent.

JeetKunDoug:
These tests were originally written expecting 1.3/1.4/2.0 for their legacy/previous/current, which means either they must updated as time goes on or be to forced to run against specific versions rather than legacy/previous/current. If trying to keep these tests working with 1.3.4 is important, we should specify exact version numbers for these tests, like others have done for 2.0.2/2.0.4, rather than depending on legacy/previous.

javajolt:
@fadushin @JeetKunDoug  That's interesting. One of the design changes I'm working on with @jburwell for Riak Test 1.1+ is the ability to tag versions with symbol names like `latest_1.3_ee` or `previous_2.0_ee`.  If tests are really that version-specific in the future we can just tag these dependencies explicitly.

PR #775, opened 2015-04-10
Potentially fix race condition in cluster join causing BTA-116 and other test failures.

JeetKunDoug:
This is a short-term fix that introduces a delay (via wait_for_service) in riak_test - need to actually investigate the underlying issue in Riak that is causing this issue (nodes transitioning from 'valid' to 'joining' and never actually joining.


macintux:
:+1:

PR #776, opened 2015-04-10
BTA-186 verify_crdt_capability

fadushin:
Added a wait for riak_kv to start before polling http interface.

This seems to address a timing issue on Linux, that does not seem to show on OS X.


JeetKunDoug:
+1

PR #778, opened 2015-04-10
update yz_handoff test for swiftness via single leave-state check... nee...

zeeshanlakhani:
...d to extend this w/ aggressive settings and more states

was failing on the builder on the **join** state. I ran into the same issue on a fresh env intermittently when using aggressive settings. Need to spend some more time w/ this and getting it working and waiting for transfers correctly on join and allowing for env speedups. Relates to YZ's extended work w/ AAE, which is coming up this week.

Handles BTA-176 **for now**.

@aberghage will review and test on the builder too, before +1'ing.

### DIFF
--- a/tests/yz_handoff.erl
+++ b/tests/yz_handoff.erl
@@ -33,9 +33,11 @@
         [
          {riak_core,
           [
-           {ring_creation_size, 16},
-           {handoff_concurrency, 11},
-           {claimant_tick, 1000}
+           {ring_creation_size, 16}
+          ]},
+         {riak_kv,
+          [
+           {handoff_rejected_max, infinity}
           ]},
          {yokozuna,
           [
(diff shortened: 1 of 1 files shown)


aberghage:
Passes locally several times in a row, looks fine to me as a stopgap.

Builders are busy doing another big test run, I'll give it a go this weekend once that wraps.

zeeshanlakhani:
@aberghage cool.

aberghage:
Still failing intermittently on the builders with FoundCount not matching ExpectedCount after the join, which eventually causes a timeout.

aberghage:
Still fails intermittently on the builders, but with lower frequency.

The failure appears to be an AAE-related race, but it's not clear whether it's a test issue or a yz/aae issue. As such, this test should probably be removed from giddyup (or minimally failures of this class ignored in the release process) until the underlying issue can be corrected.

In the meantime, this PR is fine to merge. +1 74b510b

PR #780, opened 2015-04-13
Add some extra info out to detect point of timeout in test failures

russelldb:
Pretty dumb extra logging so those intermittent giddyup failures give a better log account of themselves.


JeetKunDoug:
+1

PR #781, opened 2015-04-13
Fix test hang when riak_core_cluster_conn:status failed to respond after 2 milliseconds.

JeetKunDoug:
Should resolve test failures with a message similar to `@riak_core_cluster_conn:handle_info:402 Unmatch message {<20563.30238.10>,status}` in the riak server logs.


andrewjstone:
:+1: c603e8b

PR #783, opened 2015-04-14
Remove handling of timeout so pre-2.0 node connections will be detected

JeetKunDoug:
Old-school Riak connection nodes would also result in a timeout. Try new riak_core_cluster_conn:status() first, then fall back to older 1.4.X style bare send. Lather, rinse, repeat.


andrewjstone:
:+1: 	da28931

Great work @JeetKunDoug

PR #784, opened 2015-04-15
Add test removing core.properties and checking whether creation works ...

zeeshanlakhani:
...without issue using new *unload* check.

* Needs and tests https://github.com/basho/yokozuna/pull/487


aberghage:
+1 ac7cc9f

PR #785, opened 2015-04-15
Fix occasional failures in ensemble_sync

nickelization:
This fixes some rare race conditions in ensemble_sync where we could sometimes run wait_for_stable prior to an ensemble actually becoming unstable, and then it would pass the wait but the ensemble could become unavailable during the next step in the test.

By waiting for the ensemble leader tick counts to increment, we can guarantee that any failures will have been "noticed" prior to our calling wait_for_stable, because the leader_tick function ensures a quorum is present when it executes, and steps down if it fails to get one.

Thanks to AJS for providing consultation on this one in a number of areas :+1:


andrewjstone:
:+1: df438b9

Great work @nickelization !

nickelization:
Thanks!

PR #786, opened 2015-04-16
Riak Test v1.1.0

jburwell:
**This PR is not intended to be merged at this time.  It is being used to review the work done thus far, and begin the process of cleanup for merge into master.**

Riak Test v1.1.0 ...  more writeup to come.

/cc @javajolt


regression_test_wrapper.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
jburwell: Should we remove this script prior to merge?  We had always intended to remove it, but it may prove useful until for verifying changes until the v1 wrapper is dropped in v2.0.0.

src/giddyup.erl
```diff
@@ -27,6 +27,7 @@
 get_suite(Platform) ->
```
jburwell: Does it make sense for this module to be a gen_server?  It would be nice to remove the reliance on rt_config and pass the state information into it from the escript.  Also, it would remove potential blocking of the new pipeline if we have a slow connection to Giddyup.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
- {backend,            $b, "backend",  atom,       "backend to test [memory | bitcask | eleveldb]"},
- {upgrade_version,    $u, "upgrade",  atom,       "which version to upgrade from [ previous | legacy ]"},
- {keep,        undefined, "keep",     boolean,    "do not teardown cluster"},
- {report,             $r, "report",   string,     "you're reporting an official test run, provide platform info (e.g. ubuntu-1204-64)\nUse 'config' if you want to pull from ~/.riak_test.config"},
- {file,               $F, "file",     string,     "use the specified file instead of ~/.riak_test.config"}
+ {help,                   $h, "help",     undefined,  "Print this usage page"},
+ {config,                 $c, "conf",     string,     "specifies the project configuration"},
+ {tests,                  $t, "tests",    string,     "specifies which tests to run"},
+ {suites,                 $s, "suites",   string,     "which suites to run"},
+ {groups,                 $g, "groups",   string,     "specifiy a list of test groups to run"},
```
jburwell: Support for groups is not currently present in the planner.  Therefore, I think we should remove this option, and backlog completion of the feature.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
+ {dir,                    $d, "dir",      string,     "run all tests in the specified directory"},
+ {skip,                   $x, "skip",     string,     "list of tests to skip in a directory"},
+ {verbose,                $v, "verbose",  undefined,  "verbose output"},
+ {outdir,                 $o, "outdir",   string,     "output directory"},
+ {backend,                $b, "backend",  atom,       "backend to test [memory | bitcask | eleveldb]"},
+ {upgrade_path,           $u, "upgrade-path", atom,   "comma-separated list representing an upgrade path (e.g. riak-1.3.4,riak_ee-1.4.12,riak_ee-2.0.0)"},
```
jburwell: The new upgrade test case behaviors will not be complete for 1.1.0.  Therefore, it seems appropriate to remove this command line option, and backlog completion of this feature.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
+%% end,
+
+%% TODO: Do this in the test runner
+%%     CoverDir = rt_config:get(cover_output, "coverage"),
+%% CoverFile = rt_cover:maybe_export_coverage(Test, CoverDir, erlang:phash2(TestMetaData)),
+%% publish_report(SingleTestResult, CoverFile, Report),
+
+%%     [{coverdata, CoverFile} | SingleTestResult].
+
+%% TODO: Use `TestCount' and `Completed' to display progress output
```
jburwell: Commented code needs to be removed -- cruft bits that get in the way.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
-    %% Sets up extra paths earlier so that tests can be loadable
-    %% without needing the -d flag.
-    code:add_paths(rt_config:get(test_paths, [])),
+finalize(TestResults, Args) ->
+    %% TODO: Fixup coverage reporting
+    %% [rt_cover:maybe_import_coverage(proplists:get_value(coverdata, R)) ||
+    %%     R <- TestResults],
+    %% CoverDir = rt_config:get(cover_output, "coverage"),
+    %% Coverage = rt_cover:maybe_write_coverage(all, CoverDir),
+    %% Verbose = proplists:is_defined(verbose, Args),
```
jburwell: Commented code needs to be removed -- more crufty bits that get in the way.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
+maybe_add_code_path(Path, true) ->
+    code:add_patha(Path);
+maybe_add_code_path(_, false) ->
+    meh.
+
+load_and_start(Application) ->
+    application:load(Application),
+    application:start(Application).
+
+ensure_dir(undefined) ->
```
jburwell: This function seems more appropriately located in rt_util with some unit tests.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
-    Backends = case proplists:get_all_values(backend, ParsedArgs) of
-        [] -> [undefined];
-        Other -> Other
-    end,
-    Upgrades = case proplists:get_all_values(upgrade_version, ParsedArgs) of
-                   [] -> [undefined];
-                   UpgradeList -> UpgradeList
-               end,
+-spec comma_tokenizer(string(), [string()]) -> [string()].
+comma_tokenizer(S, Acc) ->
```
jburwell: This function seems more appropriately located in rt_util with some unit tests.

src/riak_test_escript.erl
```diff
@@ -21,299 +21,436 @@
...
-    case {kvc:value(Field, SMeta, undefined), kvc:value(Field, CMeta, undefined)} of
-        {X, X} ->
-            filter_merge_meta(SMeta, CMeta, Rest);
-        {_, undefined} ->
-            filter_merge_meta(SMeta, CMeta, Rest);
-        {undefined, X} ->
-            filter_merge_meta(lists:keystore(Field, 1, SMeta, {Field, X}), CMeta, Rest);
-        _ ->
-            false
+get_group_tests(Tests, Groups) ->
```
jburwell: Strike this function.  Group filtering should be part of the planner.

jburwell:
# Notes on the planner design

``rt_planner`` and ``rt_reporter`` are good starts towards the pipeline design I have in mind for r_t.  After reviewing the initial implementation, I have the following observations:

* The GiddyUp dependency can be further isolated in the codebase
* The riak_test_escript module contains more test planning work than necessary
* Isolate the use of rt_config to the riak_test_escript module as much as possible to simplify component unit testing

Ideally, we have a pipeline constructed with the following stages:
```
planner->scheduler->runner
```
Each of these stages can emit a test_result event that is broadcast to a list of reporters (e.g. console, giddyup).  

A planner stage accepts a list of tests, backends, and groups and creates a set of test plans to be scheduled for execution.  The following is (roughly) the test planner API I have in mind:
```
-spec start_link(pid(), [pid()], proplists:proplist() -> rt_util:result().
start_link(SchedulerPid, ReporterPids, Options) -> %% the options are implementation dependent

-spec plan([atom()], [backend()], [atom()]) -> {ok, non_neg_integer()} | rt_util:error().
plan(Tests, Backends, Groups) ->  %% When the test list is empty, run all available tests
```

A scheduler stage accepts a test plan, reserves the quantity of nodes requested by the plan, and submits the schedule to an executor stage for execution.  The following is (roughly) the test scheduler API I have in mind:

```
-spec start_link(pid(), [pid()], proplists:proplist()).
start_link(ExecutorPid, ReporterPids, Options) -> %% Options are implementation dependent

-spec schedule(#rt_planner:test_plan()) -> rt_util:result().
schedule(TestPlan) ->
```
Finally, a Runner stage provisions the nodes allocated in the test schedule, executes the test specified in the test schedule, and reports the results of the test run the to reporters.  The following is (roughly) the API I have in mind:

```
-spec start_link([pids()], proplists:proplist()) -> rt_util:result().
start_link(ReporterPids, Options) ->

-spec run(rt_scheduler:test_schedule) -> rt_util:result().
run(TestSchedule) ->
```

Based on this (high-level) design, I propose the following refactorings:

1. Convert the ``giddyup`` module to a ``gen_server`` that has the ``giddyup`` configuration information passed into its start_link function.  In addition to encapsulating the GiddyUp configuration, it will also allow asynchronous communication with GiddyUp to prevent blocking of the pipeline if/when we have a slow connection.
1. Refine the ``rt_planner`` module to define the planner behavior and any common functions used by all planners
1. Move all GiddyUp related planning functions into a new ``rt_giddyup_planner`` module -- removing the notion of wrapping test plans in the escript module.  If a list of tests, backends, and/or groups is provided to this planner, they will be used to filter the list of test plans retrieved from GiddyUp.
1. Create an ``rt_simple_planner`` module which is used when a list of tests is specified on the command line.
1. Add a build_test_pipeline function to the riak_test_escript that selects and starts the planner, scheduler, and reporter(s) and ties them together for execution.

/cc @javajolt

regression_test_wrapper.sh
javajolt: I think it's still useful for now.  Easy to remove later.

src/giddyup.erl
javajolt: Don't know how much this would give us if it were a gen_server.  We'd still end up blocking.  The problem is, we **have** to wait for GiddyUp to finish uploading all of the log files before we can clean up the node and go on.

Easy to pass in the config parameters to giddyup:get_suite/1 if you'd like.

src/riak_test_escript.erl
javajolt: I like this plan.  What about suites?

src/riak_test_escript.erl
javajolt: Yes. Same with the config options.  Still need to figure out how those will work.  We'll have to address it once we port an upgrade test.

src/riak_test_escript.erl
javajolt: Very happy to clean up.  Thought you had left it in for a reason.

src/giddyup.erl
jburwell: @brett since we are copying the logs and configs to a local results directory, there is no need to wait.  We simply give the reporter the location of the logs in the results directory.  Also, we will need to this model when we support parallelization since r_t will not be physically co-located with the Riak nodes.

src/riak_test_escript.erl
jburwell: @brett good catch -- suites need to go as well.  I created [BTA-205](https://bashoeng.atlassian.net/browse/BTA-205) to remove the suites, groups, and upgrade_path options.  I also added [BTA-203](https://bashoeng.atlassian.net/browse/BTA-203) and [BTA-204](https://bashoeng.atlassian.net/browse/BTA-204) to add back groups and command line specification of the upgrade path in a later release.

src/giddyup.erl
javajolt: @jburwell Currently the logs are *only* copied locally for a non-GiddyUp run.  If we'd like to change that, we can.  We'd also have to make sure that local logs are then periodically cleaned up the builder.  Or just decide that once the files are uploaded we simply get rid of them.

jburwell: @brett in a world where r_t is testing remote nodes, we are going to need to copy them to a local directory in order to upload them to GiddyUp.  Therefore, we should always copy the logs to the results directory.

I think it sufficient to notify @ooshlablu and @gcymbalski that r_t is storing the logs with the results locally, and they script out a reaping policy.  We should also add a note to the changes file to let users know that their local results directory may grow as well.

seancribbs: You could also invert it so the harness knows how to upload artifacts, given a test result ID. Then it could be parallelized across all the hosts in an rtssh setup, or uploaded locally in the runner for a local node.

Just a thought.

PR #779, opened 2015-04-12
YZ Extractor Test and New Helpers

zeeshanlakhani:
* add some yz wait_until helper fns
* add count_calls and helpers for tracing fn calls (with dbg)
* more yz specific helpers, helps w/ future migration of yz tests
* add yz_extractors test to check for updates made in RIAK-1673 (#382) (#382)
* add yz_noop_extractor intercept for testing a custom mimetype for extractor

Related to testing RIAK-1673 (#382) (#382) and relies on https://github.com/basho/yokozuna/pull/483.

PR #787, opened 2015-04-21
Update core properties test to make sure that indexes can be reindexed a...

zeeshanlakhani:
...fter removing of data dirs and segment info files

* Also added some of the yz/rt/yz_rt extensions in https://github.com/basho/riak_test/pull/779... this should be reviewed and merged first and I'll rebase accordingly.


zeeshanlakhani:
know that @aberghage will be reviewing this one :).

fadushin:
+1 61e3170

PR #789, opened 2015-04-24
Fullsync replication known deadlock tests.

JeetKunDoug:
Tests (repl_handoff_deadlock_aae and repl_handoff_deadlock_fullsync) to provoke deadlock causing full_sync to hang when handoff...causes vnode to exit before the fold request is started.

Requires the following PRs to pass:
https://github.com/basho/riak_repl/pull/671
https://github.com/basho/riak_kv/pull/1106


seancribbs:
+1 looks good.

PR #791, opened 2015-05-07
Hopefully fix the disconnection issue in the riak667_mixed test

seancribbs:


javajolt:
:+1: 2e7f093

seancribbs:
This passes on the builder?

javajolt:
@seancribbs This passed on my devrel version

seancribbs:
Ok, but it passed on mine too before the change. That's why I ask.

javajolt:
So this only fails on the builder?

PR #792, opened 2015-05-14
test default bucket type upgrade on change to entropy iteration... an…

zeeshanlakhani:
…d move assert cap checks to rt.erl

Related to https://github.com/basho/yokozuna/pull/492.

For this and other related tests moving forward, it'd be helpful to choose a specific version as the `previous` version to upgrade from just for that test (not on a whole config level), so that we could check that the correct updates/checks are asserted going from that version to `current`. @fadushin @jburwell @javajolt... any ideas there... is that already a thing, am i dumb? I know we can pass `-u`, but that's still choosing from the test config for all tests. Maybe I'm missing something?


javajolt:
@zeeshanlakhani After we merge our changes, the versions will have whatever label you'd like.  In fact, it would be better for many tests to specify the version.  This should be available for upgrades, too.  In the mean time, I'm not sure how you could change the versions for one test alone.  I guess you could introduce your own config parameters, but that would clutter the config file and would probably need to be removed in the future.

zeeshanlakhani:
Yeah. I'm going to clean-up the core caps unknown checks... as that's what get specifics in versions. But, yeah, that's what I figured. Thanks @javajolt.

javajolt:
So I just ran `riak_test` on test `yz_default_bucket_type_upgrade`.  Is there some other required setup?  Is it dependent upon a specific branch of YZ?
```
================ yz_default_bucket_type_upgrade failure stack trace =====================
{{badmatch,
     {error,
         <<"Invalid bucket properties: [{search_index,\n                                <<\"test_upgrade_idx does not exist\">>}]">>}},
 [{yz_rt,write_data,4,[{file,"src/yz_rt.erl"},{line,38}]},
  {yz_default_bucket_type_upgrade,confirm,0,
      [{file,"tests/yz_default_bucket_type_upgrade.erl"},{line,64}]},
=========================================================================================
```

zeeshanlakhani:
Previous should be like 2.0.5.

Zeeshan Lakhani
programmer | 
software engineer at @basho | 
org. member/founder of @papers_we_love |
twitter => @zeeshanlakhani

zeeshanlakhani:
@fadushin + @javajolt this now contains the test for the upgrade to default b_t's fix and the rename of the `yz_rt` module for test/buildbot clarity.

zeeshanlakhani:
@javajolt let me know if you're still having trouble. Make sure you're running yokozuna on https://github.com/basho/yokozuna/pull/492.

javajolt:
@zeeshanlakhani Ok, my "current" Riak for r_t is `develop` with `bugfix/zl/handle-upgrade-for-fix-to-default-bucket-type-iteration-bug` as the YZ dep. My r_t is `add-test/zl/add-test-for-default-b_t-upgrade-checks`.

zeeshanlakhani:
@javajolt hmm, odd. And, now w/ yz_rt being renamed to yokozuna_rt right? I know @fadushin ran it w/o running into that issue. I can't reproduce myself on my mac or ubuntu vm. What's your `previous` Riak though... 2.0.5?

zeeshanlakhani:
I'd make sure you've updated w/ the rename and do a `make clean` to make sure the previous yz_rt binary was removed.

javajolt:
@zeeshanlakhani Nope, it's still `yz_rt.erl` in riak_test.  Interesting.  Just updated the branch and got the renamed file.  Looks like I pulled it before your second commit.  Still getting the same error, however.  I'm running r_t directly.  Is that how you are testing it, too?
```
./riak_test -c rtdev -t yz_default_bucket_type_upgrade
```
Also, where does the `test_upgrade_idx` index get created in the test?  I see where it is being written to, but not created https://github.com/basho/riak_test/blob/add-test/zl/add-test-for-default-b_t-upgrade-checks/tests/yz_default_bucket_type_upgrade.erl#L64.  Or does the initial write create the index?

zeeshanlakhani:
@javajolt... The search index gets created in that yokozuna_rt.erl's `write_data` fun. I run r_t directly too. I'd just make sure the environment is in good shape. I'll check w/ @aberghage and @fadushin too, but it's odd. Do you see errors creating the index in the logs? The Erlang client has a sane default, which should be fine for a single index, but you can try adding the options params, `[{timeout, <timeout>}]` to the erlang client call in `write_data`. Still can't reproduce here. The error seems to be saying that the index wasn't created by the time we attempt to set it. Are you changing riakc's default ping_timeout somewhere?

zeeshanlakhani:
Using specific versions now.

```
{rt_ee, [
    {rt_project, "riak_ee"},
    {yz_dir, ["/Users/zeeshanlakhani/riak_ee/deps/yokozuna"]},
    {rtdev_path, [{root, "/Users/zeeshanlakhani/rt/riak_ee"},
                  {current, "/Users/zeeshanlakhani/rt/riak_ee/current"},
 	          {previous, "/Users/zeeshanlakhani/rt/riak_ee/riak-ee-2.0.5"},
                  {"2.0.5", "/Users/zeeshanlakhani/rt/riak_ee/riak-ee-2.0.5"},
                  {"2.1.1", "/Users/zeeshanlakhani/rt/riak_ee/riak-ee-2.1.1"},
		  {legacy, "/Users/zeeshanlakhani/rt/riak_ee/riak-ee-1.4.12"}]}
]}.
```

Still checking on issues w/ a 2.0.5 upgrade. But, working for me on the 2.1.1 to what's current upgrade.

javajolt:
Why would you need these specific versions if "current" = 2.1.1, "previous" = 2.0.5 and "legacy" = 1.4.12?

zeeshanlakhani:
@javajolt b/c as we move forward, tests would fail checking certain compatibilities and such when previous becomes 2.0.6, etc...

jburwell:
@zeeshanlakhani unless the test is for a specific upgrade condition (e.g. checking the 2.0.4 CRDT upgrade data loss bug), your test should be written to support arbitrary releases.  If the test requires a minimum version (e.g. 2.1.1), then it should check that the version being tested meets that requirement (i.e. node version >= 2.1.1).

zeeshanlakhani:
After talking to @jburwell, the thought is that I'll specify 2.0.5 w/ this test as it is the *min* version needed to test the upgrade (and cause previous would mean 2.1.1, as we're moving to 2.1.2 next @javajolt ).

In giddyup, we can specify the min_version to run this test w/, and the test will eventually be skipped once we're not dealing w/ the min version anymore.

zeeshanlakhani:
Looking more at the giddyup code, I'm guessing that previous will mean 2.0.5 for our 2.0.6 release and 2.1.1 for our 2.1.2? If that's the case, then previous is fine. Just want to be sure of that though @javajolt.

fadushin:
The only suggestion I have is to refactor verify_num_found_query into yokozuna_rt, as it is defined in two places, but there will be plenty of opportunity for that in the future.

Also, what is the purpose of the rpc:call to get the anti_entropy_build_limit environment variable for the riak_kv app?  First, I don't see any tests following this state change.  Also, I might be wrong about this, but wouldn't the riak_kv app need to be restarted in order for this change to affect behavior?

zeeshanlakhani:
@fadushin updated based on your comment :).

fadushin:
+1 ad5004c

fadushin:
+1 e86616f

PR #795, opened 2015-05-18
Don't hardcode with riak_kv.

cmeiklejohn:
Riak Core based applications, like Lasp, don't use riak_kv, and
therefore can't use riak_test if it's wiating for riak_kv to come
online.

Add support so the user can override the service the test/app is waiting
for.

Reverts breaking change in: b128e474f9961dacb6a85fcbd04078deab8f7610.


javajolt:
@cmeiklejohn Look liked #796 supersedes this, right?  If so, can we delete this branch?

cmeiklejohn:
Yes, sorry.  That is my mistake.

cmeiklejohn:
Deleted.

javajolt:
Thanks!

PR #805, opened 2015-06-26
Use correct shutdown.

cmeiklejohn:
No longer use the shutdown command when starting, which causes the use
of the incorrect version of shutdown (shutdown from current, not
previous) which generates an error.

```
{badrpc,{'EXIT',{{badfun,#Fun<erl_eval.20.80484245>},
                 [{erlang,apply,2,[]},
                  {rpc,'-handle_call_call/6-fun-0-',5,
                       [{file,"rpc.erl"},{line,203}]}]}}}
```


jonmeredith:
+1 892e8c5

Passed on my local machine with patch, woefully hung before.

PR #806, opened 2015-06-26
Make hwm 250 for easier debugging

mrallen1:
This change sets lager to only drop messages from error_logger if they exceed 250 messages per second.  It should help us not lose messages when trying to debug some kind of messy application start up issue.


jonmeredith:
+1 7fa4f3a

PR #808, opened 2015-07-07
update some yokozuna tests from work done in riak/2.0 branch for 2.0.…

zeeshanlakhani:
…6 release

@fadushin, @macintux, @JeetKunDoug please take a look. We decided to stick w/ a specific, previous version last I remember us discussing this for 2.0.6 (so we don't have to rely on `previous`). If so, we can re-run these tests, and I can make an update on giddyup in preparation for the next release.

### DIFF
--- a/tests/yz_default_bucket_type_upgrade.erl
+++ b/tests/yz_default_bucket_type_upgrade.erl
@@ -17,6 +17,13 @@
 %% under the License.
 %%
 %%--------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


fadushin:
Re-tested against 2.1.2pre1 and both tests run fine.  Not sure what I was doing wrong on my 2.1 branch, other than not using locked deps.  Strange.

+1 b68ffbc

PR #809, opened 2015-07-07
update yokozuna crdt test from master that works for current 2.0 bran…

zeeshanlakhani:
…ch and 2.0.6 release


fadushin:
Enriched crdt test passes on Linux.

+1 cf0b665

PR #811, opened 2015-07-14
Add test for sink rebalance code

bsparrow435:
This test adds a check to make sure a 2 node cluster can balance sink connections with another 2 node cluster. It also checks for a known `badrecord` issue. The fix for this issue is https://github.com/basho/riak_repl/pull/667

### DIFF
--- a/tests/repl_rt_sink_rebalance.erl
+++ b/tests/repl_rt_sink_rebalance.erl
@@ -81,7 +81,7 @@ confirm() ->
     ?assertEqual(ok, rt:wait_until_all_members(BNodes)),
     ?assertEqual(ok, rt:wait_until_ring_converged(BNodes)),
     rt:wait_until(AFirst, fun(Node) ->
-        Whereis = rpc:call(Node, erlang, whreis, [riak_core_connection_manager]),
+        Whereis = rpc:call(Node, erlang, whereis, [riak_core_connection_manager]),
         is_pid(Whereis)
     end),
     repl_util:wait_for_connection(AFirst, "B"),
--- a/tests/repl_rt_sink_rebalance.erl
+++ b/tests/repl_rt_sink_rebalance.erl
@@ -76,14 +76,15 @@ confirm() ->
 
     rt:stop(BFirst),
     ?assertEqual(ok, rt:wait_until_unpingable(BFirst)),
-    lager:info("Let sleep for 10s"),
-    timer:sleep(10000),
     rt:start(BFirst),
     ?assertEqual(ok, rt:wait_until_nodes_ready([BFirst])),
     ?assertEqual(ok, rt:wait_until_all_members(BNodes)),
     ?assertEqual(ok, rt:wait_until_ring_converged(BNodes)),
-    timer:sleep(10000),
-    lager:info("Sleeping again to allow realtime repl to connect"),
+    rt:wait_until(AFirst, fun(Node) ->
+        Whereis = rpc:call(Node, erlang, whreis, [riak_core_connection_manager]),
+        is_pid(Whereis)
+    end),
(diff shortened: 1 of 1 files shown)


tests/repl_rt_sink_rebalance.erl
```diff
@@ -0,0 +1,137 @@
...
+    connect_clusters(AFirst, BFirst),
+
+    enable_rt(AFirst, ANodes),
+
+    [?assertEqual(ok,verify_connectivity(Node)) || Node <- ANodes],
+    [?assertEqual(ok,verify_sinks(Node)) || Node <- BNodes],
+
+    rt:stop(BFirst),
+    ?assertEqual(ok, rt:wait_until_unpingable(BFirst)),
+    lager:info("Let sleep for 10s"),
```
JeetKunDoug: Can we avoid sleeping for 10s here? For example, check if cluster manager is missing the IP of the node?

tests/repl_rt_sink_rebalance.erl
```diff
@@ -0,0 +1,137 @@
...
+    timer:sleep(10000),
+    rt:start(BFirst),
+    ?assertEqual(ok, rt:wait_until_nodes_ready([BFirst])),
+    ?assertEqual(ok, rt:wait_until_all_members(BNodes)),
+    ?assertEqual(ok, rt:wait_until_ring_converged(BNodes)),
+    timer:sleep(10000),
```
JeetKunDoug: Given verify_sinks uses rt:wait_until, can we drop the timer:sleep here?

tests/repl_rt_sink_rebalance.erl
```diff
@@ -0,0 +1,137 @@
...
+    lager:info("Sleeping again to allow realtime repl to connect"),
+
+    [?assertEqual(ok,verify_sinks(Node)) || Node <- BNodes],
+    [check_for_badrecord(Node) || Node <- ANodes],
```
JeetKunDoug: I think verify_sinks is probably sufficient to test for this issue - it seems that check_for_badrecord is so generic ("error.*{badrecord,state}*" could happen for many reasons having nothing to do with this test, for example) that it's not a good way to validate the fix, and could fail later for a completely unrelated reason.

tests/repl_rt_sink_rebalance.erl
bsparrow435: So maybe my understanding is a bit out of date but riak_test used to have a stop issue where without waiting you had a chance of starting before the node was down causing issues. If this is no longer the case, we can happily remove it.

tests/repl_rt_sink_rebalance.erl
bsparrow435: So the issue here is we run verify_sinks too fast and look for the connection_manager before it exists resulting in a badrpc for reason noproc. This crashes the test. I should have changed verify sinks to also wait until the cluster_manager exists instead of the sleep but I knew that this would work while I was testing.

tests/repl_rt_sink_rebalance.erl
bsparrow435: So the issue that inspired this test is actually a specific repl issue where we throw this badrecord error[1][2] because of a bad noreply in an fsm callback(hooray lack of checking).

The fact that this fails with verify_sinks sometimes is a totally different issue where the reconnect code is not balancing correctly, even in this simple two 2-node cluster case. We could make the lager check more specific regex but I couldnt identify a way to check for this crash outside of lager since the process in questions is restarting very frequently and clearing the incorrect state on resurrection.

[1] https://github.com/basho/riak_repl/issues/666
[2] https://github.com/basho/riak_repl/pull/667

bsparrow435:
thanks for updating @lordnull ! Looks ready to merge

PR #812, opened 2015-07-16
Test fix for bad_match during handoff with write_once buckets

fadushin:
This PR fixes the verify_handoff_write_once test to reliably exercise the riak_kv_vnode:handle_handoff_command, which is fixed under

https://github.com/basho/riak_kv/pull/1153

The test achieves this goal in inserting a new intercept into the riak_kv_worker which will force the test to write data into riak while handoff is running (but before the vnode is notified that handoff is complete).  This should address any race conditions that existed before this fix.


JeetKunDoug:
:+1: 7e7cb73

PR #813, opened 2015-07-17
Add basic cluster converge test to try to capture issues with cluster convergence outside of other tests.

JeetKunDoug:


javajolt:
:+1: 1e39711

PR #814, opened 2015-07-20
Add verify_snmp_repl riak_test

jvoegele:
The verify_snmp_repl test reproduces the error condition (one cluster
configured for realtime replication with two or more other clusters) and
passes if and only if the riak_snmp_stat_poller module correctly handles
this configuration.

https://bashoeng.atlassian.net/browse/RIAK-1884 (#23)
https://github.com/basho/riak_snmp/issues/27


JeetKunDoug:
TODO: Let's see if post-2.1.2 we can make a more generic "test SNMP polling with MDC" sort of test with different cluster topologies.

PR #816, opened 2015-07-24
Fix verify_membackend test

javajolt:
Adjust `verify_membackend:get_used_space/2` to work with Riak 2.1+.  Three issues had been causing it to fail:
1. The regex in `get_used_space` did not recognize `riak*-2.1`
2. The structure of the state of the vnode had changed since 2.0
3. The size of Riak Object had changed so the expected amount of memory used had changed

### DIFF
--- a/tests/verify_membackend.erl
+++ b/tests/verify_membackend.erl
@@ -152,7 +152,10 @@ check_put_delete(Node) ->
     Mem = get_used_space(Pid, Node),
 
     %% this is meh, but the value isn't always the same length.
-    case (Mem == MemBaseline - 1142) orelse 
+    %% It seems to be the size of a Riak Object
+    case (Mem == MemBaseline - 1146) orelse
+        (Mem == MemBaseline - 1145) orelse
+        (Mem == MemBaseline - 1142) orelse
         (Mem == MemBaseline - 1141) of
         true ->
             ok;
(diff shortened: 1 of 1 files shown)


tests/verify_membackend.erl
```diff
@@ -152,7 +152,10 @@ check_put_delete(Node) ->
     Mem = get_used_space(Pid, Node),
 
     %% this is meh, but the value isn't always the same length.
-    case (Mem == MemBaseline - 1142) orelse 
+    %% It seems to be the size of a Riak Object
+    case (Mem == MemBaseline - 1146) orelse
```
mrallen1: I wonder if this test would be less brittle if it just set some kind of upper threshold like "Baseline - 1150" and made sure that Mem > Threshold or something like that.  Maybe there's some reason the test isn't already like that?

tests/verify_membackend.erl
```diff
@@ -252,35 +255,55 @@ get_remote_vnode_pid(Node) ->
                                all_vnodes, [riak_kv_vnode]),
     VNode.
 
-%% this is silly fragile
+%% @doc Crack open the VNode state record to find
+%% the hidden number of used bytes within the
+%% riak_kv_memory_backend:state record.
+-spec parse_regular_state_fun(integer()) -> fun().
+parse_regular_state_fun(Offset) ->
```
mrallen1: Is there a better more generalized way to handle this?  For example, maybe we could have a private function call in the memory backend that either returned the memory usage?

tests/verify_membackend.erl
```diff
@@ -252,35 +255,55 @@ get_remote_vnode_pid(Node) ->
...
+    fun(X) ->
+        element(
+            3, lists:nth(
+                1, element(
+                    2, element(
+                        Offset, element(
+                            4, element(2, X))))))
+    end.
+
+%% this is silly fragile and only works for Riak 2.0+
```
mrallen1: Same comment as above - I think the solution might be to make the backends support a reporting mechanism rather than trying to parse some opaque data object.

tests/verify_membackend.erl
javajolt: Not sure why the test was set up this way. I suspect it was the size of the Riak Object, but of course it can change over time.  I could set a range, but as soon as the size changes, the test will break again.

tests/verify_membackend.erl
javajolt: I agree completely, which I think should be the next stage for this test.  Right now I'm trying to get it to pass with 2.1.  Having the memory backend report its own usage sounds like a great idea.

tests/verify_membackend.erl
mrallen1: Yeah.  Well. What is the point of checking this anyway? To make sure there's not some kind of memory leak?

mrallen1: It makes sure that if a delete happens, memory is reclaimed.  So one thing might be to grab the riak object using the Key and ask the system what size it is and then make sure when its deleted, the memory drops by that amount.

tests/verify_membackend.erl
JeetKunDoug: Yea, the parsing of state records like that is pretty horrible - I guess for 2.0/2.1 we could just add the "private tell me your memory usage" function, and keep the horribleness for 1.4 until we officially kill it. That said, I'm OK with this fix for now if the test runs for 2.1, and we can continue to clean it up post-2.1.2 release as part of the "flappy test rewrite" work.

javajolt: @JeetKunDoug I'm looking at where to add the code in riak_kv to have it parse the state symbolically instead of a horribly brittle mess it is now.  I'm hopeful that part will not be too bad.

PR #815, opened 2015-07-21
Update Documentation and Build Scripts for Riak 2.1

javajolt:
Use the documented scripts now to build
- 1.4.12
- 2.0.2 (needed for specific tests)
- 2.0.4 (needed for specific tests)
- 2.0.6 
- current (develop)

PR #817, opened 2015-07-26
Fix riak_control and riak_control_authentication for Riak 2.1+

javajolt:
- Use the actual version string (e.g. riak_ee-2.1.2) of devrel
  instead of "current", "previous", "legacy" to determine which
  tests to run
- Add rt:get_version/1 to find the version of devrels other than
  "current"

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -72,6 +72,7 @@
          get_replica/5,
          get_ring/1,
          get_version/0,
+         get_version/1,
          heal/1,
          http_url/1,
          https_url/1,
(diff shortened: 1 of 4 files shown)


javajolt:
@mrallen1 Let me know if you've had a chance to test this so we can knock it off the list for 2.1.2

mrallen1:
@javajolt I'm going to spend tomorrow working on all your open PRs so we can get them merged.

cmeiklejohn:
Does this need a review?

javajolt:
Thanks, @cmeiklejohn, but @mrallen1 is looking at it already.  I'll assign to him to avoid confusion.

mrallen1:
We got sidetracked by #815 @cmeiklejohn - but thanks for the offer :cake:

mrallen1:
:+1: f349c02

jonmeredith:
@javajolt does this need to land on riak/2.0 as well?

javajolt:
@jonmeredith This should work for riak/2.0, too.  It should work for an older configuration of current=2.0, previous=1.4, etc as well as current=2.1, previous=2.0, legacy=1.4.

javajolt:
@jonmeredith Not sure how you are using the riak/2.0 branch, but it would work fine with an older current=2.0, previous=1.4, legacy=1.3 setup, but the test only broke with the introduction of 2.1.

jonmeredith:
The riak/2.0 branch is what gets checked out for running 2.0 builds against
on the builders.
The master branch is run for 2.1 tests.

PR #818, opened 2015-07-30
Allow secondary location for intercepts. 

cmeiklejohn:
Allow the configuration to specify an alternative location for intercept files, instead of assuming that they are located with riak_test/intercepts; this allows projects to keep related intercepts locally.

### DIFF
--- a/src/rt_intercept.erl
+++ b/src/rt_intercept.erl
@@ -25,11 +25,12 @@
 files_to_mods(Files) ->
     [list_to_atom(filename:basename(F, ".erl")) || F <- Files].
 
-default_intercept_path_glob() ->
-    filename:join([rt_local:home_dir(), "intercepts", "*.erl"]).
+default_intercept_path_globs() ->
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
cc: @macintux

src/rt_config.erl
```diff
@@ -51,7 +51,7 @@ get_os_env(Var, Default) ->
 
 %% @doc Load the configuration from the specified config file.
 load(Config, undefined) ->
-    load(Config, filename:join([os:getenv("HOME"), ".riak_test.config"]));
+    load(Config, filename:join([os:getenv("PWD"), ".riak_test.config"]));
```
macintux: I understand the motivation, but shouldn't this be additive? Check `pwd` first, if it's not there, check `HOME`?

cmeiklejohn: This change shouldn't have made it in; this is a Lasp-specific change.

macintux:
Once you've reverted the PWD/HOME bit, this looks good.

PR #819, opened 2015-07-30
If we're building on OS X, don't build HiPE or ODBC connector

mrallen1:
These cause OS X to fail (after the build in HiPE's case, and during the build in ODBC case).  So just skip them.

### DIFF
--- a/bin/rtdev-build-releases.sh
+++ b/bin/rtdev-build-releases.sh
@@ -55,8 +55,12 @@ kerl()
     RELEASE=$1
     BUILDNAME=$2
 
+    if [[ $(uname -s) -eq "Darwin" ]]; then
+        export KERL_CONFIGURE_OPTIONS='--without-hipe --without-odbc'
+    fi
+
     echo " - Building Erlang $RELEASE (this could take a while)"
-    ./kerl build $RELEASE $BUILDNAME  > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


mrallen1:
@javajolt Please consider merging this into #815

javajolt:
A little suggestion while we are tweaking:
`--enable-darwin-64bit` seems liked a good addition for OSX and `--enable-m64-build` everywhere else.  Also maybe we should just disable hipe and odbc by default?  I found these defaults in Jared's build script for creating Basho™ Erlang:
`--disable-hipe --enable-smp-support --without-odbc`

mrallen1:
I dunno. HiPE seems fine on Linux. But it kills OS X in high load / high memory scenarios.

javajolt:
Also dug this on the interwebs:
```
env MAKE="make -j10" kerl build R16B03
```
Maybe we can add that to use more than one core in the builds

javajolt:
As mentioned in HipChat, here are the official scripts for [R15](https://github.com/basho/basho_buildbot/blob/master/erlscripts/install_patched_r15.sh) and [R16](https://github.com/basho/basho_buildbot/blob/master/erlscripts/install_patched_r16.sh)

mrallen1:
@javajolt OK, I added all of the build flags from the official scripts and the make -j10 flag which really seems to boost things a good deal. I think this is ready for a merge.

javajolt:
:+1: Let's merge for now

PR #821, opened 2015-07-31
Update verify_memorybackend for RIAK-2061

javajolt:
Previously this test had a very brittle function, `get_used_space/2` to decompose the state of a VNode.  The problem was that any time the structure of the `#state` record changed, the test would fail.  I had updated this to work with 2.1 in https://github.com/basho/riak_test/pull/816, but realized that in subsequent releases we'd be back in the same spot.  So instead I've added functions to KV and Core to expose parts of their states needed to determine the underlying memory used.

`check_put_delete/1` also had some brittle magic numbers in there which essentially is the size of the Riak Object written. Sadly size of the object going into `riakc` is very different from what the memory backend actually receives due to the updated metadata.  Instead of guessing the size of the object written, I added [put_obj_size](https://github.com/basho/riak_kv/blob/bugfix/bch/cleanup_verify_membackend/src/riak_kv_memory_backend.erl#L89) into the memory backend state so we can track the size of the last object written.  It's arguable that it's a waste of space, but it should only be a few bytes and will (hopefully) make this crusty test more reliable.

N.B.: This PR requires both https://github.com/basho/riak_kv/pull/1163 and https://github.com/basho/riak_core/pull/758 to work successfully.  These are on the 2.0 branch and will need to be merged forward.  This conglomeration was tested against `riak_ee-2.0.6`.

### DIFF
--- a/tests/verify_membackend.erl
+++ b/tests/verify_membackend.erl
@@ -7,14 +7,6 @@
 
 -define(BUCKET, <<"ttl_test">>).
 
-%% from 2.0, but should be valid for 1.1+
--record(state, {data_ref :: ets:tid(),
-                index_ref :: ets:tid(),
-                time_ref :: ets:tid(),
-                max_memory :: undefined | integer(),
-                used_memory=0 :: integer(),
-                ttl :: integer()}).
-
 confirm() ->
     Tests = [ttl, max_memory, combo],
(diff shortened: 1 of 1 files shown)


mrallen1:
:+1: 4034777

PR #823, opened 2015-08-04
Tweak rtdev-build-releases.sh to use the correct Erlangs

javajolt:
The logic to determine which version of Erlang to build was incorrect

### DIFF
--- a/bin/rtdev-build-releases.sh
+++ b/bin/rtdev-build-releases.sh
@@ -81,9 +81,9 @@ kerl()
 
     echo " - Building Erlang $RELEASE (this could take a while)"
     # Use the Basho-patched version of Erlang 
-    if [ "$RELEASE" == "R15B01" ]; then
+    if [ ! -z $(echo "$RELEASE" | grep "R15B01") ]; then
         BUILD_CMD="./kerl build git git://github.com/basho/otp.git $R15_TAG $BUILDNAME"
-    elif [ "$RELEASE" == "R16B02" ]; then
+    elif [ ! -z $(echo "$RELEASE" | grep "R16B02") ]; then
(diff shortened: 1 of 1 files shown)


bin/rtdev-build-releases.sh
```diff
@@ -81,9 +81,9 @@ kerl()
 
     echo " - Building Erlang $RELEASE (this could take a while)"
     # Use the Basho-patched version of Erlang 
-    if [ "$RELEASE" == "R15B01" ]; then
+    if [ ! -z $(echo "$RELEASE" | grep "R15B01") ]; then
         BUILD_CMD="./kerl build git git://github.com/basho/otp.git $R15_TAG $BUILDNAME"
-    elif [ "$RELEASE" == "R16B02" ]; then
+    elif [ ! -z $(echo "$RELEASE" | grep "R16B02") ]; then
```
mrallen1: I don't understand the `! -z ...` Why not just write it like

```bash
case "$RELEASE" in
   "R15B01")
     BUILD_CMD="..."
     ;;
   "R16B02")
```
and so on...

mrallen1: This seemed to work before anyway.  Here the output of bash -x bin/rtdev-build-releases.sh
```bash
echo ' - Building Erlang R16B02 (this could take a while)'
 - Building Erlang R16B02 (this could take a while)
+ '[' R16B02 == R15B01 ']'
+ '[' R16B02 == R16B02 ']'
+ BUILD_CMD='./kerl build git git://github.com/basho/otp.git OTP_R16B02_basho9 erlang-R16B02'
```

mrallen1: Now the bash operator for string equality is actually `-eq` but it still worked with `==`

mrallen1:
I'm not sure this is needed. It may be that the particular problem build is only found on OS X.

PR #826, opened 2015-08-11
Bugfix/jdm/cluster does not converge race

macintux:
Two changes:
* Test for race condition identified in riak_core#764
* Use new error tuple created in riak_core#764 to recognize premature join and retry

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -328,6 +328,9 @@ deploy_nodes(Versions, Services) ->
                                               Service <- Services ],
     Nodes.
 
+deploy_nodes(NumNodes, InitialConfig, Services) when is_list(hd(InitialConfig)), is_integer(NumNodes) ->
+    NodeConfig = [{current, Config} || Config <- InitialConfig],
+    deploy_nodes(NodeConfig, Services);
 deploy_nodes(NumNodes, InitialConfig, Services) when is_integer(NumNodes) ->
     NodeConfig = [{current, InitialConfig} || _ <- lists:seq(1,NumNodes)],
     deploy_nodes(NodeConfig, Services).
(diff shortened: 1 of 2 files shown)


macintux:
https://github.com/basho/riak_core/pull/764 since I can't get the Markdown correct

mrallen1:
I know @cmeiklejohn is reviewing, but this is good work. Thanks John.

javajolt:
@macintux This is very exciting to (probably) have this long-standing issue fixed.

cmeiklejohn:
Why wasn't ```wait_until``` used here?  It already takes the default timeout into account when attempting to perform an action...

src/rt.erl
```diff
@@ -411,9 +411,19 @@ join(Node, PNode) ->
...
+    timer:sleep(500),
+    retry_if_join_error(Fun(), Fun, RetryCount - 1);
+retry_if_join_error(ok, _Fun, _RetryCount) ->
+    ok.
+
 %% @doc Have `Node' send a join request to `PNode'
 staged_join(Node, PNode) ->
-    R = rpc:call(Node, riak_core, staged_join, [PNode]),
+    F = fun() -> rpc:call(Node, riak_core, staged_join, [PNode]) end,
+    R = retry_if_join_error(F(), F, 5),
```
cmeiklejohn: Why not combine these two into a ```wait_until```?

macintux:
You're right, @cmeiklejohn, `wait_until` is the right choice. Will rework.

macintux:
Ready for another look. I verified the use of `wait_until` by calling `rt:staged_join` from the new test and verifying that the retry occurred as expected.

javajolt:
```
Test Results:
verify_build_cluster_caps_race-bitcask: pass
0 Tests Failed
1 Tests Passed
```

javajolt:
:+1: d3b5956

PR #827, opened 2015-08-12
Quickcheck test for vnode proxy overload recovery.

jonmeredith:
Requires vnode proxy fix (probably release *after* 1.4.12/2.0.6/2.1.2).

Test for https://github.com/basho/riak_core/pull/767 (RIAK-1988)

### DIFF
--- a/tests/proxy_overload_recovery.erl
+++ b/tests/proxy_overload_recovery.erl
@@ -15,7 +15,7 @@
 %%
 %% Whenever the vnode is resumed, the test checks the message queues are
 %% cleared and the model is reset.  The main goal is to advance the proxy
-%% into interestign new states.
+%% into interesting new states.
 %%
 %% This test can be run outside of riak_test while working on it. 
 %% Symlink the source into a release build and run
@@ -260,11 +260,15 @@ resume_args(#tstate{rt = RT}) ->
 
 resume(#rt{ppid = PPid, vpid = VPid}) ->
     sys:resume(VPid),
+    %% Use the sys:get_status call to force a synchronous call
+    %% against the vnode proxy to ensure all messages sent by
+    %% this process have been serviced and there are no pending
+    %% 'ping's in the vnode before we continue.
+    %% Then drain the vnode to make sure any pending pongs have
+    %% been sent.
     ok = drain(VPid),
-    %% make sure all messages have been processed and there
-    %% is no pending 'ping' in the vnode before we continue.
-    sys:get_status(PPid),
-    sys:get_status(VPid).
+    _ = sys:get_status(VPid),
+    _ = sys:get_status(PPid).
 
 resume_next(S, _V, _A) ->
     S#tstate{vnode_running = true, proxy_msgs = 0, direct_msgs = 0}.
@@ -317,7 +321,7 @@ overloaded_args(#tstate{vnode_running = Running, rt = RT}) ->
 overloaded(Running, #rt{ppid = PPid, vpid = VPid}) ->
     case Running of
         true ->
-            ok = drain(PPid), % make sure all proxy msgs procesed/dropped
+            ok = drain(PPid), % make sure all proxy msgs processed/dropped
             ok = drain(VPid); % make sure any pending ping/pongs are processed
         _ ->
             ok
@@ -383,7 +387,8 @@ prep_env(Var, Val) ->
 
 %%
 %% Wait until all messages are drained by the Pid. No guarantees
-%% about future messages being sent.
+%% about future messages being sent, or that responses for the
+%% last message consumed have been transmitted.
 %%
 drain(Pid) ->
     case erlang:process_info(Pid, message_queue_len) of
@@ -426,17 +431,16 @@ setup_eqc(Nodes) ->
 
 %% Add the QuickCheck in the r_t runner to the list of Nodes
 add_eqc_apps(Nodes) ->
-    Apps = [eqc, pulse, pulse_otp],
-    [begin
-         %% Check if the remote node has EQC installed
-         case rpc:call(Node, code, priv_dir, [eqc]) of
-             %% If not, add EQC to embedded node from rt host. 
-             %% Assume same filesystem. Sorry rtssh.
-             {error, bad_name} ->
-                 [rpc:call(Node, code, add_pathz,
-                           [code:lib_dir(App, ebin)], 60000) || App <- Apps];
-             Path when is_list(Path) ->
-                 ok
-         end
-     end || Node <- Nodes],
+    Apps = [eqc, pulse, pulse_otp, eqc_mcerlang],
+    [case code:lib_dir(App, ebin) of
+         {error, bad_name} -> % EQC component not installed locally
+             ok;
+         Path when is_list(Path) ->
+             case rpc:call(Node, code, priv_dir, [App]) of
+                 {error, bad_name} ->
+                     true = rpc:call(Node, code, add_pathz, [Path], 60000);
+                 _ ->
+                     ok
+             end
+     end || App <- Apps, Node <- Nodes],
     ok.
--- a/tests/proxy_overload_recovery.erl
+++ b/tests/proxy_overload_recovery.erl
@@ -25,6 +25,10 @@
 %% On failure you can re-run counter examples *and* print out the internal
 %% state with the run.
 %%   proxy_overload_recovery:check(). 
+%%
+%% TODO/Questions:
+%% 1) Is there a better way to do the initialization step?
+%% 2) Remove the riak_kv dependency so it's a pure riak_core test.
 
 -module(proxy_overload_recovery).
 -include_lib("eqc/include/eqc.hrl").
--- a/tests/proxy_overload_recovery.erl
+++ b/tests/proxy_overload_recovery.erl
@@ -42,15 +42,23 @@ confirm() ->
     Opts = [{riak_kv, [{anti_entropy, {off, []}}]}],
     Nodes = [Node1] = rt:deploy_nodes(1, Opts),
 
-    %% Distribute this code to the remote and add EQC
-    rt:load_modules_on_nodes([?MODULE], Nodes),
-    add_eqc_apps(Nodes),
+    case setup_eqc(Nodes) of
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
What's the proper way to run this test?

cmeiklejohn:
I have a valid EQC license, but I'm not sure how to get this test to run correctly.

```
================ proxy_overload_recovery failure stack trace =====================
{{assertEqual_failed,
     [{module,proxy_overload_recovery},
      {line,52},
      {expression,"rpc : call ( Node1 , ? MODULE , rtrun , [ TestingTime ] )"},
      {expected,true},
      {value,
          {badrpc,
              {'EXIT',
                  {licence_authorization_failed,eqc,
                      {licence_authorization_failed,eqc_licence,
                          {{wrong_code_file_versions,
                               [eqc_mcerlang,eqc_mce_prop_transform,
                                eqc_mcerlang_statem]},
                           [{eqc_licence,f733768_0,1,
                                [{file,"../src/eqc_licence.erl"},
                                 {line,60}]}]}}}}}}]},
 [{proxy_overload_recovery,'-confirm/0-fun-1-',3,
      [{file,"tests/proxy_overload_recovery.erl"},{line,52}]},
  {proxy_overload_recovery,confirm,0,
      [{file,"tests/proxy_overload_recovery.erl"},{line,52}]},
==================================================================================

16:56:27.085 [notice] proxy_overload_recovery Test Run Complete
16:56:27.086 [info] Stopping cover

Test Results:
proxy_overload_recovery-bitcask: fail
1 Tests Failed
0 Tests Passed
That's 0.0% for those keeping score


So, we find ourselves in a tricky situation here. 
You've run a single test, and it has failed.
Would you like to leave Riak running in order to debug?
[Y/n] 
```

jonmeredith:
Yikes.  It magically runs for me and sets the EQC paths.  Until I resolve the rt issues you can just compile inside a running node and just call the run(Seconds) function.

jonmeredith:
@cmeiklejohn it should use the EQC you have available now.

cmeiklejohn:
Doesn't appear to be working.

```
================ proxy_overload_recovery failure stack trace =====================
{{assertEqual_failed,
     [{module,proxy_overload_recovery},
      {line,53},
      {expression,"rpc : call ( Node1 , ? MODULE , rtrun , [ TestingTime ] )"},
      {expected,true},
      {value,
          {badrpc,
              {'EXIT',
                  {licence_authorization_failed,eqc,
                      {licence_authorization_failed,eqc_licence,
                          {{wrong_code_file_versions,
                               [eqc_mcerlang,eqc_mce_prop_transform,
                                eqc_mcerlang_statem]},
                           [{eqc_licence,f733768_0,1,
                                [{file,"../src/eqc_licence.erl"},
                                 {line,60}]}]}}}}}}]},
 [{proxy_overload_recovery,'-confirm/0-fun-1-',3,
      [{file,"tests/proxy_overload_recovery.erl"},{line,53}]},
  {proxy_overload_recovery,confirm,0,
      [{file,"tests/proxy_overload_recovery.erl"},{line,53}]},
==================================================================================

19:59:12.131 [notice] proxy_overload_recovery Test Run Complete
19:59:12.139 [info] Stopping cover
```

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+%%   1) If the vnode message queue has cleared, the vnode proxy should not
+%%   be in overload.
+%%   2) If the proxy has been sent more messages than the overload threshold
+%%      while the vnode is suspended, it should be in overload.
+%%   3) If the proxy has been sent less than or equal to the overload threshold
+%%      it should not be in overload.
+%%
+%% Whenever the vnode is resumed, the test checks the message queues are
+%% cleared and the model is reset.  The main goal is to advance the proxy
+%% into interestign new states.
```
cmeiklejohn: Typo 'interestign'

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+-include_lib("eqc/include/eqc.hrl").
+-include_lib("eqc/include/eqc_statem.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-export([confirm/0]).
+-compile(export_all).
+
+confirm() ->
+    %% disable AAE to prevent vnode start delay opening AAE leveldb
+    Opts = [{riak_kv, [{anti_entropy, {off, []}}]}],
+    Nodes = [Node1] = rt:deploy_nodes(1, Opts),
```
cmeiklejohn: Most of the new tests uses the `build_clusters` API, and no longer call `deploy_nodes` directly.

jonmeredith: What's the best way to deploy a singleton, that's all this test needs.

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+
+%%
+%% Run prepare once before any other calls and then never again.
+%%
+precondition_common(#tstate{rt = undefined}, {call, _M, F, _A}) ->
+    F == prepare;
+precondition_common(_, {call, _M, F, _A}) ->
+    F /= prepare.
+
+%% %% Make sure we're still running what we think we're running - uncomment
```
cmeiklejohn: Maybe remove this dead code?

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+    catch
+        _:Err ->
+            lager:info("EQC unavailable: ~p\n", [Err]),
+            {error, unavailable}
+    end.
+
+
+%% Add the QuickCheck in the r_t runner to the list of Nodes
+add_eqc_apps(Nodes) ->
+    Apps = [eqc, pulse, pulse_otp],
```
cmeiklejohn: Should we commit the `eqc_mcerlang` change?

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+        {'DOWN', Mref, _, _, _} ->
+            ok
+    end.
+
+%%
+%% Check if EQC is available on the local node, then add to remotes if needed
+%% This could be refactored out if many tests need it.
+setup_eqc(Nodes) ->
+    try
+        _ = eqc:reserved_until(), % will throw exception if problems
```
cmeiklejohn: Do the builders have eqc?  If not, will this return as a test failure?

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+
+resume_args(#tstate{rt = RT}) ->
+    [RT].
+
+resume(#rt{ppid = PPid, vpid = VPid}) ->
+    sys:resume(VPid),
+    ok = drain(VPid),
+    %% make sure all messages have been processed and there
+    %% is no pending 'ping' in the vnode before we continue.
+    sys:get_status(PPid),
```
cmeiklejohn: You're calling `get_status` but doing nothing with the response...?

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+    sys:get_status(VPid).
```
cmeiklejohn: Similiar... shouldn't you verify the queue has been drained?

tests/proxy_overload_recovery.erl
```diff
@@ -0,0 +1,442 @@
...
+%%
+%% Check the current state of the vnode so we can test properties
+%%
+overloaded_args(#tstate{vnode_running = Running, rt = RT}) ->
+    [Running, RT].
+
+overloaded(Running, #rt{ppid = PPid, vpid = VPid}) ->
+    case Running of
+        true ->
+            ok = drain(PPid), % make sure all proxy msgs procesed/dropped
```
cmeiklejohn: Spelling `procesed`

cmeiklejohn:
:+1: d43e526

Some minor nitpicking, but a cursory read of the code looks good and is straightforward; model simply randomly sends direct messages vs. proxy messages and verifies that the overload code compensates accordingly.

tests/proxy_overload_recovery.erl
jonmeredith: Which lines do you think are dead? 

The precondition match on rt = undefined makes sure it is always the only task that can be selected first. The second clause makes sure that prepare is never selected again once the runtime parameters are created.

tests/proxy_overload_recovery.erl
jonmeredith: Yes, I need to fix it, but probably make it a better check to see if it exists locally.

tests/proxy_overload_recovery.erl
jonmeredith: The intent was that if eqc was not installed, or the license could not be reserved then we issue a warning saying the test did not actually run and pass it.

Unfortunately there's no way of skipping that I could work out looking at the r_ sources.

tests/proxy_overload_recovery.erl
jonmeredith: Yes, it's a trick to make sure a synchronous call goes through the proxy pid without changing any state, guaranteeing any messages that have been sent will be processed before the next commands.  I'll clarify the comment further.

tests/proxy_overload_recovery.erl
jonmeredith: The thought was - make sure the vnode is drained, which may send pending pong messages back to the proxy. Drain does not guarantee there were not messages pending, so the get_status checks make sure that happens. 

However, I think the order is wrong - this is an implementations specific detail which is why a pulse test would be better, but as the proxy does not respond to pong messages, we should ensure they are sent by the vnode *before* we check there are no pending messages being processed in the proxy.

Nice catch.

jonmeredith:
@cmeiklejohn thanks for the review - fixed things up and pushed.

cmeiklejohn:
:+1: 865a389

Nicely done.

cmeiklejohn:
Does this need to be merged into the mainline as well?  What about GiddyUp?

jonmeredith:
Yes. Will merge it forward and add.

PR #832, opened 2015-08-13
Make this test deterministic whether it likes it or not: intercept init

macintux:
Blame @jonmeredith


jonmeredith:
+1 this is the best evil I've seen all week.

PR #835, opened 2015-08-14
wrap contents of tests/proxy_overload_recovery in -ifdef(EQC)

hmmr:
andalso, provide the missing `-behaviour(riak_test).`.


macintux:
@hmmr if you'll re-file this against `riak/2.0` branch, I'll plus it; we're trying to apply the same logic to rt as other projects.

PR #842, opened 2015-08-21
Create new connections after they crash, rather than depending on auto_reconect

JeetKunDoug:
The Erlang client's auto_reconnect on our builders seems to not work quite correctly - this change forces a new connection after each known point that the server-side of the connection would crash, to hopefully resolve the {error, disconnected} issues we see on our build infrastructure.

### DIFF
--- a/tests/riak667_mixed.erl
+++ b/tests/riak667_mixed.erl
@@ -49,8 +49,8 @@ confirm() ->
     lager:info("mdc_crdt_epoch? ~p", [rpc:multicall(Nodes, application, get_env, [riak_kv, mdc_crdt_epoch])]),
 
     %% Create PB connection.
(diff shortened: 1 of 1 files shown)


fadushin:
Runs cleanly on my CentOS machine.
+1 94c66e7

javajolt:
So you guys just gave up on the reconnect and just created a new connection each time?  I guess debugging riak_pb is not really the aim of this test.  However, seems like we should create a test on riak_pb to see why reconnecting is not working.

JeetKunDoug:
Yea - So we'll add an issue on the riak client to try to track down where auto disconnect is failing us, but this test, as you pointed out, really has nothing to do with that, so getting it to pass reliably was more important.

PR #847, opened 2015-08-25
Added rt:create_activate_and_wait_for_bucket_type

JeetKunDoug:
Allows tests to wait for all nodes in the cluster to have the new type before continuing. Update http_bucket_types to use said function.


fadushin:
Paired with @JeetKunDoug on this.

+1 31b1002

javajolt:
Did we need this for the 2.0 branch, too?

JeetKunDoug:
Sadly, yes. 2.0 for some reason doesn't fail as often, but I did (now that you mention it) find it in a 2.0.6 build. Will backport.

PR #850, opened 2015-08-25
Feature/vk/spark master test

zkhadikova:


tests/bdp_service_manager.erl
```diff
@@ -0,0 +1,139 @@
...
+%%      combinations of starting, stopping and execution nodes
+
+-module(bdp_service_manager).
+-behavior(riak_test).
+-export([confirm/0]).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(S1_NAME, "service-one").
+-define(S1_TYPE, "cache-proxy").
+%% (this is parameter 'Group' in various calls to data_platform_global_state functions)
```
lordnull: Comment seems a bit misleading. It's referring tot he line above, rather than the line below. It also seems to indicate the macro is used as the group name _instead of_ rather than _in addition to_ the service type.

tests/bdp_spark.erl
```diff
@@ -0,0 +1,138 @@
...
+-behavior(riak_test).
+-export([confirm/0]).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(SERVICE_1, "spark-service-one").
+-define(SERVICE_2, "spark-service-two").
+-define(SPARK_MASTER_TYPE, "spark-master").
+-define(SPARK_IDENT_STRING, "bdp_spark_test").
+-define(HOSTNAME, "test").
```
lordnull: The name of the machine I'm using is not 'test', but 'ubuntu-tester'. This hard-coded assumption causes test failures.

tests/bdp_spark.erl
```diff
@@ -0,0 +1,138 @@
...
+-module(bdp_spark).
+-behavior(riak_test).
+-export([confirm/0]).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(SERVICE_1, "spark-service-one").
+-define(SERVICE_2, "spark-service-two").
+-define(SPARK_MASTER_TYPE, "spark-master").
+-define(SPARK_IDENT_STRING, "bdp_spark_test").
```
lordnull: The ident string seemed to be ignored (or had no effect since my machine has no "bdp_spark_test" user in the spark environment, so it ran as my user. This hard-coded assumption caused test failures.

tests/bdp_spark.erl
```diff
@@ -0,0 +1,138 @@
...
+    %% 15/07/09 02:13:46 INFO Master: I have been elected leader! New state: RECOVERING
+
+    [DevPath | _] = rtdev:devpaths(),
+
+    Node1Path = DevPath ++ "/dev/dev" ++ integer_to_list(1),
+    Node2Path = DevPath ++ "/dev/dev" ++ integer_to_list(2),
+
+    lager:info("Node1 path = ~s, Node2 path = ~s", [Node1Path, Node2Path]),
+
+    LogSubPath = "/lib/data_platform-1/priv/spark-master/logs/spark-" ++ ?SPARK_IDENT_STRING ++ "-org.apache.spark.deploy.master.Master-1-" ++ ?HOSTNAME ++  ".out",
```
lordnull: When a devrel is built, there's no guarantee that the data_platform directory will be "data_platform-1". If stagedevrel is uses, the directory is just "data_platform". Once we starting using real versions, it will be "data_platform-$vsn"; further indicating that this would need to be updated for each time data_platform's version changes. Using "data_platform*" should work.

tests/bdp_util.erl
```diff
@@ -0,0 +1,221 @@
...
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%% @doc Helper functions for bdp_service_manager.erl and others.
+
+-module(bdp_util).
+
+-export([build_cluster/1, build_cluster/2,
+         service_added/4, service_removed/2, service_started/4, service_stopped/4,
```
lordnull: These 4 function names seem misleading. The names indicate checking something that happened in the past or a current state, but they are actually attempting to do an action. I'd recommend the following changes to adhere to "least surprise":

* service_added -> add_service
* service_removed -> remove_service
* service_started -> start_service
* service_stopped -> stop_service

tests/bdp_util.erl
```diff
@@ -0,0 +1,221 @@
...
+%% is returned immediately, which is weird, but goes away after some 8
+%% seconds of trying.  It's reasonable to stick that to SM itself and
+%% get it fixed there; for the time being, this workaround is here.
+
+call_with_patience(Node, M, F, A) ->
+    call_with_patience_(Node, M, F, A, ?SM_RPC_RETRIES).
+call_with_patience_(_Node, _M, _F, _A, 0) ->
+    lager:error("Exhausted ~b retries for an RPC call to ~p ~p:~p/~b",
+                [?SM_RPC_RETRIES, _Node, _M, _F, length(_A)]),
+    ?assert(false);
```
lordnull: Rather than just using an assert false that give no real indication of what when wrong, throwing a descriptive error will have a clearer failure:

     error({retries_exhausted, {Node, M, F, A}});

tests/bdp_util.erl
```diff
@@ -0,0 +1,221 @@
...
+call_with_patience_(Node, M, F, A, Retries) ->
+    case rpc:call(Node, M, F, A) of
+        {badrpc, Reason} ->
+            lager:error("RPC call to ~p failed with reason: ~p", [Node, Reason]),
+            ?assert(false);
```
lordnull: An error with the badrpc is much clearer than the assert.

tests/bdp_util.erl
```diff
@@ -0,0 +1,221 @@
...
+    Rnn1 = lists:usort(Rnn0 -- [ServiceName]),
+    ok = wait_services(Node, {Rnn1, Avl0}).
+
+
+
+%% Makeshift SM API: node join/leave.
+%% Bits of code copied here from data_platform_console
+%% This really calls for implementing proper SM API in data_platform_*.
+
+%% TODO: make behave
```
lordnull: What does "make behave" mean?

tests/bdp_spark.erl
```diff
@@ -0,0 +1,138 @@
...
+
+    lager:info("Spark service one log path ~s", [Spark1LogFile]),
+    lager:info("Spark service two log path ~s", [Spark2LogFile]),
+ 
+    os:cmd("chmod +x ./priv/bdp_spark_test/leader_election_check.sh"),
+    Command = "./priv/bdp_spark_test/leader_election_check.sh " ++ Spark1LogFile ++ " " ++ Spark2LogFile ++ " " ++ ?TIMEOUT,
+    lager:info("Running bash script: ~s", [Command]),
+    Res = os:cmd(Command),
+    lager:info("Res= ~s", [Res]),
+    ?assert(Res == "ok\n"),
```
lordnull: Using `?assertEquals("ok\n", Res)` instead of a raw assert means if the assert fails, it will print both the expected and received values. This means the log statement above is unneeded; in the passing case we know what it is and don't need it output, and in the failing case, the value is output.

lordnull:
I was unable to the bdp_spark test to pass; it seems to be that I used stagedevrel to create my dev's. This means all my data_platform priv directories in each devrel point to the same place. This may be something that can be changed in the node config's when setting up the cluster. It may also be something that can't be resolved, and this test cannot be run on a stagedevrel'ed setup.

tests/bdp_util.erl
hmmr: That's my half-assed attempt to remove and then re-add a node. I couldn't get it to work (more importantly, couldn't make up my mind on what the 'removal' should be -- make node leave BDP ensemble? leave the riak cluster? down the node outright?).

hmmr:
This branch should probably not be merged into master, as the dbp_spark and related modules can only work with data_platform_ee flavour of riak.

tests/bdp_util.erl
hmmr: This function properly belongs to the data_platform_core_ee (where is exists in https://github.com/basho/data_platform_core_ee/pull/57, in https://github.com/basho/data_platform_core_ee/commit/d09d233b589519331a84c39a12449216c9b944bf).

hmmr:
@lordnull I addressed the comments you made in dbp_util, and @zkhadikova has fixed (some) of the other ones. When we are done, we will close this PR and open another, against feature/az/bdp.

And, concerning issues you have had running it on stagedevrel, it's not going to work because spark wants to write to priv/ (which is bad in itself, but that's outside the scope at the moment). You do need a devrel build.

tests/bdp_spark.erl
zkhadikova: Machine name is used by default, unless we provide value, which we do. So this shouldn't be an issue.

tests/bdp_spark.erl
zkhadikova: Current user name is used by default, which is not the case. We need to override default values so that we can define log files.

PR #851, opened 2015-08-25
Allow devrel name in .riak_test.config to be an atom or a string

javajolt:
Also add `rt_util.erl` with various utilities developed earlier this year with John Burwell:
- `spec find_atom_or_string(term(), orddict:orddict()) -> term() | undefined.`
- `spec convert_to_string(string()|atom()) -> string().`
- `spec convert_to_atom(string()|atom()) -> atom().`
- `spec convert_to_atom_list(atom()|string()) -> undefined | list().`
- `spec pmap(F :: fun(), L :: list()) -> list().` - To me removed from `rt.erl` in another PR

Running `riak667_mixed` test vs 2.0 is a good way to test this.  Especially with some mixed release names, like "2.0.2" and '2.0.4'.  Of course legacy, previous and current are already atoms.

Relates to [CI-61](https://bashoeng.atlassian.net/browse/CI-61)

### DIFF
--- a/src/rt_util.erl
+++ b/src/rt_util.erl
@@ -1,0 +1,141 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(rt_util).
+-include_lib("eunit/include/eunit.hrl").
+
+-export([convert_to_atom/1,
+         convert_to_atom_list/1,
+         convert_to_string/1,
+         find_atom_or_string/2,
+         find_atom_or_string_dict/2,
+         pmap/2]).
+
+%% @doc Look up values by both atom and by string
+find_atom_or_string(Key, Table) ->
+    case {Key, proplists:get_value(Key, Table)} of
+        {_, undefined} when is_atom(Key) ->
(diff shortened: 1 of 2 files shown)


src/rtdev.erl
```diff
@@ -764,3 +764,20 @@ get_node_logs() ->
           {ok, Port} = file:open(Filename, [read, binary]),
           {lists:nthtail(RootLen, Filename), Port}
       end || Filename <- filelib:wildcard(Root ++ "/*/dev/dev*/log/*") ].
+
+-ifdef(TEST).
+
+release_versions_test() ->
+    ok = rt_config:set(rtdev_path, [{root, "/Users/hazen/dev/rt/riak"},
```
hmmr: If you use `os:getenv("RT_DEST_DIR")` in place of hardcoded path, all Linux users will say thank you (as well as all those whose login is not 'hazen':)

src/rt_util.erl
```diff
@@ -0,0 +1,141 @@
...
+        true ->
+            [list_to_atom(X) || X <- string:tokens(Values, ", ")];
+        _ ->
+            [list_to_atom(Values)]
+    end.
+
+%% @doc Parallel Map: Runs function F for each item in list L, then
+%%      returns the list of results
+-spec pmap(F :: fun(), L :: list()) -> list().
+pmap(F, L) ->
```
hmmr: There seems to be a `pmap/2` function in `rt`: https://github.com/basho/riak_test/blob/master/src/rt.erl#L1733. (and even a second one in observer.erl).

fadushin:
rt_util:pmap/2 is inert, but will be used in the future.  We should investigate whether pmap is intended to preserve the order of outputs.  If so, we can use something like:

    pmap(F, L) ->
        Self = self(),
        Pids = [spawn(fun() -> Self ! {self(), F(E)} end) || E <- L],
        [receive {Pid, Res} -> Res end || Pid <- Pids].

+1 ffadf64

src/rt_util.erl
javajolt: @hmmr The plan is to consolidate them in another PR.

src/rtdev.erl
javajolt: @hmmr This is a unit test.  We could use your name, if you'd like. :wink:

hmmr:
As an acting pedant of the day, I have to note that `pmap` with a `spawn_link` can bring down the calling process, while one with a simple `spawn` can, if the argument function freaks out and generates exceptions, just never return. Here's my gist with pmap variations: https://gist.github.com/hmmr/b536f55d229dcf15b048.

PR #852, opened 2015-08-25
Fail by crashing test process instead of returning a tagged tuple

jvoegele:
`riak_test_runner` does not support tagged tuples for failures. Fix verify_snmp_repl module so that it crashes when it fails, and write more detailed error messages to the log instead.

### DIFF
--- a/tests/verify_snmp_repl.erl
+++ b/tests/verify_snmp_repl.erl
@@ -68,14 +68,15 @@ wait_for_snmp_stat_poller() ->
             pass;
         {fail, Reason} ->
             lager:error("Failure in wait_for_snmp_stat_poller: ~p~n", [Reason]),
-            fail;
+            error({fail, Reason});
         X ->
             lager:error("Unknown failure in wait_for_snmp_stat_poller: ~p~n", [X]),
-            fail
+            error(X)
     after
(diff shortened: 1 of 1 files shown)


javajolt:
So your change does allow for a more graceful crash:
```
19:27:03.887 [info] Wait until ring converged on ['dev3@127.0.0.1']
19:27:04.159 [error] Failure in wait_for_snmp_stat_poller: {error,function_clause}
19:27:04.159 [warning] verify_snmp_repl failed: fail
19:27:04.159 [error]
================ verify_snmp_repl failure stack trace =====================
fail
===========================================================================
```

jvoegele:
@javajolt Thanks for testing and for the feedback. My uncertainty is: what is the "correct" way for a riak_test test to fail? Should it return the atom `fail`? Should it call `exit` or `error` explicitly? Should it just use EUnit's `?assert*` macros?

javajolt:
@jvoegele Where are your changes to fix repl stats polling?

javajolt:
@jvoegele Yes, I had been testing against `develop` instead of 2.1.2rc3. OOops.

JeetKunDoug:
:+1:

PR #853, opened 2015-08-26
Use wait_for and actually check the results of `riak-admin cluster join`

JeetKunDoug:
to ensure we've really joined the cluster before we continue the test.


javajolt:
Paired with @JeetKunDoug and it looks good.
:+1:

PR #854, opened 2015-08-26
Forward merge/dr/riak 2.0 to master

JeetKunDoug:
Merge branch 'riak/2.0' into forward_merge/dr/riak_2.0_to_master
- Forward merge verify_stage_clustering fix
- Updates to allow either strings or atoms for node names in config & allow them to be different (atom or string) in test vs. config.
- http_bucket_types backport, which is a no-op since it's already here.


javajolt:
:+1:

PR #855, opened 2015-08-27
spark master test

hmmr:
A spark test implementing scenario I.A (https://docs.google.com/document/d/1Ry492DaJUIyeppCsOmDeVUMy63Fcmqg97MnouP_ks94/edit?pli=1#heading=h.v8mw5fchr6mm).

Re-filed against this branch as it is not intended for the general riak. For older comments and discussion, see #850).


tests/bdp_spark_worker.erl
```diff
@@ -0,0 +1,142 @@
...
+    Spark2LogFile = Node2Path ++ LogSubPath,
+
+    lager:info("Spark service one log path ~s", [Spark1LogFile]),
+    lager:info("Spark service two log path ~s", [Spark2LogFile]),
+ 
+    os:cmd("chmod +x ./priv/bdp_spark_test/leader_election_check.sh"),
+    Command = "./priv/bdp_spark_test/worker_failover.sh " ++ Spark1LogFile ++ " " ++ Spark2LogFile ++ " " ++ ?TIMEOUT,
+    lager:info("Running bash script: ~s", [Command]),
+    Res = os:cmd(Command),
+    ?assert(Res == "ok\n"),
```
lordnull: While debugging a different problem, I had to change this to `?assertEqual("ok\n", Res).`. It will provide what was expected and what was gotten on a failure (as opposed to "it was just false"), and no output on a success (as it does now).

lordnull:
Sorry for the delay; this test [exposed a bug concerning spark logging(https://github.com/basho/data_platform_core_ee/pull/61), so had to get around that before I could work on this properly. Once that pr is pulled in, we will be able to properly evaluate this pr.

vktk:
@hmmr , Regarding double 'e' in the bp_util:start_seervice/4 function. Please check the line 
https://github.com/basho/riak_test/blob/feature/az/bdp/tests/bdp_util.erl#L25
As you can see the method was renamed in bdp_util.erl itself (commit is here - https://github.com/basho/riak_test/commit/6360af1f506e6706edad63eb8528f57ef50182b1#diff-42b639306ec57e9a37400e73bdfc25f2R25), so I'm absolutely sure our code is compilable.
Do you want us to fix this typo?

hmmr:
@vktk The typo is in my code, indeed. My bad.

Then I suggest you ignore this comment and just go with double 'e'. Once this PR is accepted, I (or someone else) will fix the typo, in bdp_util.erl as well as everywhere else where it appears, in a separate PR.

zkhadikova:
@hmmr , we need timeout to give spark nodes time to start properly, elect leader and register worker before we run shell script. If there is any better way to  do this please let me know.

hmmr:
@zkhadikova Is there a way to determine the readiness of spark by grepping its logs, just like you do for the main part of the test?

PR #869, opened 2015-09-18
Add riak_test for spatial queries that use JTS

jvoegele:
This test defines a Solr schema that uses the JtsSpatialContextFactory
provided by JTS, and issues a query that uses features provided by JTS
to ensure that the inclusion of the JTS JAR file enables spatial queries
in Yokozuna.


jvoegele:
This should have been created against `master` instead of `riak/2.0`, so recreating.

PR #872, opened 2015-09-29
update

korry8911:


tests/bdp_spark_pi_submit_test.erl
```diff
@@ -0,0 +1,110 @@
...
+        bdp_util:build_cluster(
+          ClusterSize, [{lager, [{handlers, [{file, "console.log"}, {level, debug}] }]}]),
+   
+    %% add spark master and worker services
+    lager:info("Adding Spaker Master and Spark Worker Service..."), 
+    %% add master service
+    ok = bdp_util:add_service(Node1, ?SPARK_MASTER_SERVICE_NAME, ?SPARK_MASTER_SERVICE_TYPE, ?SPARK_MASTER_SERVICE_CONFIG),
+    ok = bdp_util:wait_services(Node1, {[], [?SPARK_MASTER_SERVICE_NAME]}),
+    %% add worker service
+    SparkMasterIP = re:replace(rpc:call(Node1,os,cmd,["ipconfig getifaddr en0"]),"\\s+", "", [global,{return,list}]),
```
hmmr: In the test environment, all Erlang nodes coexist on one physical node. 127.0.0.1 would then be just fine. (And, depending on the underlying host to have ipconfig installed is risky, as is the assumption that the first network interface is called "en0" rather than "eth0".)

tests/bdp_spark_pi_submit_test.erl
```diff
@@ -0,0 +1,110 @@
...
+%spark worker definitions
+-define(SPARK_WORKER_SERVICE_NAME, "my-spark-worker").
+-define(SPARK_WORKER_SERVICE_TYPE, "spark-worker").
+-define(SPARK_WORKER_SERVICE_CONFIG, [{"MASTER_URL", "spark://127.0.0.1:7077"},{"SPARK_WORKER_PORT","8081"}]).
+
+
+%this code tests if a spark job can successfully be submitted to a two node spark cluster within a three node bdp cluster
+
+%WARNINGS: spark worker configs must contain exact ip address of spark master, not 127.0.0.1.  Otherwise, spark master and worker
+%will fail to link together.  As such, os:cmd is used to get the ip address of node1 via a shell command that might not work on systems other than moc osx.
```
hmmr: These lines are way too long.

tests/bdp_spark_pi_submit_test.erl
```diff
@@ -0,0 +1,155 @@
...
+-define(SPARK_WORKER_SERVICE_TYPE, "spark-worker").
+-define(SPARK_WORKER_SERVICE_CONFIG, [{"MASTER_URL", "spark://127.0.0.1:7077"},{"SPARK_WORKER_PORT","8081"}]).
+
+
+% this code tests if a spark job can successfully be submitted to a two
+% node spark cluster within a three node bdp cluster
+
+% WARNINGS: spark worker configs must contain exact ip address of spark
+% master, not 127.0.0.1.  Otherwise, spark master and worker will fail
+% to link together.  As such, os:cmd is used to get the ip address of
```
paegun: The comment re: method of getting the spark master's ip is out of date. The code uses inet:getifaddrs .

tests/bdp_spark_pi_submit_test.erl
```diff
@@ -0,0 +1,155 @@
...
+              fun({Iface, Details}) ->
+                      case is_routed_addr(Details) of
+                          undefined ->
+                              false;
+                          Addr ->
+                              {true, {Iface, Addr}}
+                      end
+              end,
+              Ifaces);
+        {error, PosixCode} ->
```
paegun: fail fast, if we cannot reach and run inet:getifaddrs on a node, the test results should be a fail due to improper setup. Letting this failure continue makes finding the issue harder.

hmmr:
Addressed reviewer's issues. Andalso, replaced `os:cmd("wget")` with `httpc:request`.

PR #873, opened 2015-09-30
TS put test fixes

andytill:
Fixes the fail tests by asserting they return an error tuple but not the error message at this stage. We need the client to support the error codes to do that, if we wanted to.

Deleted ts_A_put_fail_3.erl since it only failed on a local function clause, it is not an integrated test.


hmmr:
+1

PR #874, opened 2015-09-30
Fix Expected value in ts_A_select_pass module

hmmr:
Note: it naively, but correctly, expects 10 records back from the select query, and thus will begin to pass once that off-by-one issue is properly fixed.


hmmr:
#875 appears to supersede this PR, so closing unmerged.

PR #875, opened 2015-10-01
Fixed select pass tests

andytill:
Match results of the selection, from ok to {Cols, Rows} where the expected Rows are exclusive the first and last key i.e. `i > 1` instead of `i >= 1'.


hmmr:
Closing #874 in favor of this PR.

PR #880, opened 2015-10-07
Feature/more ts tests

gordonguthrie:


gordonguthrie:
I'm a tube - mis-commit

PR #888, opened 2015-10-12
Parallel extract merge

macintux:
Add tests for 2i parallel extract and return_body.


rebar.config
```diff
@@ -11,11 +11,12 @@
 {eunit_opts, [verbose]}.
 
 {deps, [
+        {riak_pb, ".*", {git, "git://github.com/basho/riak_pb", {branch, "feature/jrd/index_body"}}},
```
mrallen1: Just want to make sure these are the right branches for the dependencies.

rebar.config
```diff
@@ -11,11 +11,12 @@
...
+        {riakc, ".*", {git, "git://github.com/basho/riak-erlang-client", {branch, "feature/jrd/index_body"}}},
```
mrallen1: We talked about some other branches in HipChat :)

macintux: Yep, I didn’t review rebar.config changes, will have to tweak these, thanks.

-John

mrallen1:
Looks good here. Ran these tests for 6+ hours without issue.

mrallen1:
:+1: f2cae65

PR #889, opened 2015-10-14
add a check for single key delete in ts_basic

hmmr:
RTS-201. To use with https://github.com/basho/riak-erlang-client/tree/feature/az/delete-api-call, https://github.com/basho/riak_kv/tree/feature/az/delete-api-call and https://github.com/basho/riak_pb/tree/feature/az/ts-delete.


hmmr:
Wrong target branch, sorry. Will reopen against riak_ts.

PR #896, opened 2015-10-22
include a riakc_ts:get check in ts_basic

hmmr:
RTS-202 (Single-key reads for TS). Depends on https://github.com/basho/riak_pb/pull/143.

An 'official' test module (ts_A_get_pass_1) will be added separately.


nickelization:
Built everything and ran the test out of the box, and it failed due to not using eleveldb by default, but that's a problem separate from this PR. I'll create another PR to fix that. Other than that, looks good!

nickelization:
+1 02ec602

PR #897, opened 2015-10-22
Fix overload test

JeetKunDoug:
Previously, overload had been written using "fudge factors" on top of thresholds/messages sent. As time went on, the test was made less strict, until it no longer really tested anything, but still failed on occasion. Several things have been done in this PR to fix these issues:
1) Only test things that make sense. For example, consistent gets don't spawn FSMs, so testing them in the fsm_protection test makes no sense.
2) Each test now overwrites _all_ config values, as only setting what it thought it needed was causing tests to fail in combination, but pass in isolation. Later work will split this test into several independent tests.
3) Use knowledge of the system, like Sidejob's calculated worker_limit and width, to calculate more exact expected values.
4) Use {{Type, Bucket}, Key, Value} in pattern matches rather than combinations of that and `ConsistentType` boolean by extracting BKVs into macros to be used across the test.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -604,7 +604,8 @@ load_modules_on_nodes([Module | MoreModules], Nodes)
   when is_list(Nodes) ->
     case code:get_object_code(Module) of
         {Module, Bin, File} ->
-            {_, []} = rpc:multicall(Nodes, code, load_binary, [Module, File, Bin]);
+            {ResList, []} = rpc:multicall(Nodes, code, load_binary, [Module, File, Bin]),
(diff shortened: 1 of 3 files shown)


javajolt:
I'm looking forward to taking the baby out for a spin

javajolt:
Looks great, @JeetKunDoug! I love some of the optimizations you added.  I had a few concerns about logging all of the results, but I see the original test did that so it's fine.  I tested on 2.1.2rc3 and 2.0.6 and both pass.  Let's get this committed then merge forward!

javajolt:
:+1: 818645e

mrallen1:
Wow, nice work! :cake:

PR #899, opened 2015-10-23
Force eleveldb backend for ts_basic test

nickelization:
This was crashing out of the box for me with a cryptic error about a "bad_qid". Turned out to be caused by using the wrong backend. This fix should prevent others from running into the same issue in the future.


nickelization:
Closing this - @macintux opened a superset of this PR with #904.

PR #906, opened 2015-10-23
Added preliminary test for filtering API (full test will have to wait…

erikleitch:
… until riak_ql is fixed for boolean/any data types)

This PR just adds two additional files to riak_test/tests: 

1) ts_api.erl, which contains a single large test of the API (internally, many tests are run against the same keyset, but only one riak_test has been added)

2) ts_api_util.erl, which contains utilities for constructing the api queries, constructing the list of expected records for each query, and verifying that the returned records match.  

I'm sure it's terrible erlang, but it gets the job done for testing...

### DIFF
--- a/tests/ts_api.erl
+++ b/tests/ts_api.erl
@@ -1,0 +1,95 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2015 Basho Technologies, Inc.
+%%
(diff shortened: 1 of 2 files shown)


macintux:
Our tests are filled with terrible Erlang, so you're in good company. Checking.

macintux:
:+1:

PR #908, opened 2015-10-27
include an int field and put big numbers in it

hmmr:
In ts_basic, add an int field and put some big numbers in it, both less than and greater than sint64_max, specifically to test the conversion going in riak_pb/riak_pb_ts_codec for such values (involving binary<->float/integer tricks).


hmmr:
Not applicable as no floats don't get converted through binaries anymore.

PR #910, opened 2015-10-27
Another 2 tests fixes

gordonguthrie:
The test ts_A_select_fail_1 was shadowing ts_A_put_fail_1 as you
can't get to the select if the put fails

The error message for ts_A_put_fail_3_bad_date has been added


tests/ts_A_select_fail_1.erl
```diff
@@ -1,22 +0,0 @@
--module(ts_A_select_fail_1).
-
```
javajolt: So by fix here, you just mean get rid of the test?

javajolt:
:+1:

PR #911, opened 2015-10-28
riakc_ts:get now returns a tuple

hmmr:
RTS-480. Uses API as of https://github.com/basho/riak_kv/pull/1243 (and related PRs).

### DIFF
--- a/tests/ts_basic.erl
+++ b/tests/ts_basic.erl
@@ -95,12 +95,12 @@ confirm() ->
     GetKey = [GetTimepoint, GetSensor],
     ResGet = riakc_ts:get(C, ?BUCKET, GetKey, []),
     io:format("Get a single record: ~p\n", [ResGet]),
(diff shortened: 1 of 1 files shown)


lordnull:
The test fails without the correct pr's in the other modules, and passes with them.

Failure with only r_t updated; failure with only r_t and it's associated deps updated, success when r_t, it's deps, and the riak cluster under test is updated.

I do feel I must note that in order to get to the assert added, I had to disable previous asserts that were failing. I don't consider this a blocking issue, just noting for posterity.

basho/riak-erlang-client#242, basho/riak_pb#146, and basho/riak_kv#1243 all also have clean, readable code, thus getting a +1 from me; however it should be noted that there are other concerns about riak_pb#146 (not from me) that could very well superceede.

javajolt:
:+1:

PR #912, opened 2015-10-29
Removed write_once tests

fadushin:
DO NOT MERGE FORWARD

The write_once tests and intercepts were added to riak_test before the riak/2.0 branch was created.  Write_once is a 2.1 feature, so these tests are being removed from the riak/2.0 branch.


javajolt:
@fadushin I just checked in GiddyUp and I see that the minimum version to run those tests is 2.1.0, so we should be safe on that front.

javajolt:
:+1: c682fb1

PR #917, opened 2015-11-02
Fixed ts_api test to pass once the badarith bugfix is checked in

erikleitch:
ts_api needed to have its primary key construction reversed to match the reversal of the PK construction in the DDL compiler

### DIFF
--- a/tests/ts_api.erl
+++ b/tests/ts_api.erl
@@ -99,3 +99,4 @@ confirm_NeqOps(C) ->
     ts_api_util:confirm_Pass(C, {myfloat, float,   '!=', 2.0}),
     ts_api_util:confirm_Pass(C, {mybin,   varchar, '!=', test2}),
     ts_api_util:confirm_Pass(C, {mybool,  boolean, '!=', true}).
+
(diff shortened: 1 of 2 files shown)


javajolt:
@erikleitch Do you have a link to the badarith bug?

erikleitch:
This fixes the key order, but note that the test will continue to fail until @macintux gets a PR in for the badarith bug in riak_kv_qry_compiler.erl that we found this morning

erikleitch:
@javajolt This is the fix that @macintux was working on: https://github.com/basho/riak_kv/compare/end-to-end/timeseries...bugfix/jrd/badarith.  Basically it's a a simple change to an anonymous function that gets passed to lists:fold to determine the min/max time from a set of start/end primary key.  It's assuming that the time portion of the PK lives in the first part of the start/end key tuples, which is backwards, now that the primary key order has been reversed.

javajolt:
@erikleitch @macintux's fix is now in a PR: https://github.com/basho/riak_kv/pull/1253

javajolt:
```
Test Results:
ts_api-eleveldb: pass
0 Tests Failed
1 Tests Passed
```

javajolt:
:+1:

PR #919, opened 2015-11-04
Fixed ts_A_put_pass_2 to pass.  Primary key was backwards, type 'any'…

erikleitch:
… no longer supported, booleans need to be booleans (ie, 'true') and not binaries (<<"true">>), and test must return 'pass' atom after assertions have succeeded


javajolt:
```
Test Results:
ts_A_put_pass_2-eleveldb: pass
0 Tests Failed
1 Tests Passed
```

javajolt:
:+1:

PR #930, opened 2015-11-16
Updated to attempt to make sure we get the "right" vnode PID.

JeetKunDoug:


jonmeredith:
+1 ce8ef794cb83bcff0030c330515696947774b187

PR #931, opened 2015-11-18
Refactor rt_cascading to improve stability

hazen:
This test has been very flappy in our testing.  @lordnull has greatly refactored this test to increase both clarity and stability.

### DIFF
--- a/tests/rt_cascading.erl
+++ b/tests/rt_cascading.erl
@@ -1259,16 +1259,6 @@ timeout(MultiplyBy) ->
             N * MultiplyBy
     end.
 
-maybe_skip_teardown(TearDownFun) ->
-    fun(Arg) ->
-        case rt_config:config_or_os_env(skip_teardowns, undefined) of
-            undefined ->
-                TearDownFun(Arg);
-            _ ->
-                ok
-        end
-    end.
-
 wait_until_pending_count_zero(Nodes) ->
     WaitFun = fun() ->
         {Statuses, _} =  rpc:multicall(Nodes, riak_repl2_rtq, status, []),
--- a/tests/rt_cascading.erl
+++ b/tests/rt_cascading.erl
@@ -46,6 +46,7 @@ confirm() ->
     circle(),
     pyramid(),
     diamond(),
+    circle_and_spurs(),
 
     case eunit:test(?MODULE, [verbose]) of
         ok ->
@@ -508,7 +509,27 @@ diamond_tests(Nodes) ->
         Eval()
     end, Tests).
 
-circle_and_spurs_test_() ->
+circle_and_spurs() ->
+    State = circle_and_spurs_setup(),
+    _ = circle_and_spurs_tests(State),
+    circle_and_spurs_teardown(State).
+
+circle_and_spurs_setup() ->
+    Config = [
+        {"north", 1, ["east", "north_spur"]},
+        {"east", 1, ["west", "east_spur"]},
+        {"west", 1, ["north", "west_spur"]},
+        {"north_spur", 1},
+        {"east_spur", 1},
+        {"west_spur", 1}
+    ],
+    Clusters = make_clusters(Config),
+    lists:flatten([Nodes || {_, Nodes} <- Clusters]).
+
+circle_and_spurs_teardown(Nodes) ->
+    rt:clean_cluster(Nodes).
+
+circle_and_spurs_tests(Nodes) ->
     %                        +------------+
     %                        | north_spur |
     %                        +------------+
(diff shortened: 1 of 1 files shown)


javajolt:
Tests are passing against the `riak_ts` branch.  More comments pending.

tests/rt_cascading.erl
```diff
@@ -96,7 +104,7 @@ simple_test_() ->
             repl_util:start_realtime(State#simple_state.middle, "end")
         end},
 
-        {"cascade a put from beginning down to ending", timeout, timeout(25), fun() ->
```
javajolt: @lordnull I see that you removed most (all?) of the test timeouts.  Was that one of the causes of failure?  Do we have to worry now about the tests blocking?

JeetKunDoug:
This should resolve #935  - would like for it to get merged onto 2.0/2.1/develop as well.

javajolt:
@JeetKunDoug Would you like to be a second reviewer? Meredith wanted more eyes on it since it's a particularly problematic (and complicated) test historically.

nickelization:
Also @javajolt had a good question in the review comments about timeouts being removed. Given that we've seen this test time out in the past, I'm wondering if that's a change we really should be merging in, unless @lordnull has a really good reason for it.

JeetKunDoug:
The problem is those timeouts were implemented via eunit, which can't
safely be used inside riak_test due to the fact that eunit spawns a process
to run the test, which doesn't die when the process running eunit (in this
case, the riak_test runner) dies. Because of this, the tiak_test can think
it times out but leave the eunit process running and doing all kinds of
interesting things to the system under test.

That said, perhaps we should find a more appropriate way to put those
timeouts back in?

tests/rt_cascading.erl
```diff
@@ -106,7 +114,7 @@ simple_test_() ->
             ?assertEqual(Bin, maybe_eventually_exists(State#simple_state.ending, <<"objects">>, Bin))
```
mrallen1: `<<objects>>` should be replaced with `?bucket` to match the other statements.

tests/rt_cascading.erl
```diff
@@ -1204,25 +1266,6 @@ timeout(MultiplyBy) ->
...
-            undefined ->
-                TearDownFun(Arg);
-            _ ->
-                ok
-        end
-    end.
-
 wait_until_pending_count_zero(Nodes) ->
     WaitFun = fun() ->
         {Statuses, _} =  rpc:multicall(Nodes, riak_repl2_rtq, status, []),
```
mrallen1: I guess I can only comment on the diff parts - in line 1273 below we could change this whole expression to be

```erlang
%% somewhere else...
all_true(true) -> true;
all_true(_) -> false.

%% then at line 1273
lists:all(fun all_true/1, Out)
``` 

instead of what's there.

tests/rt_cascading.erl
```diff
@@ -128,14 +136,45 @@ simple_test_() ->
...
-
-big_circle_test_() ->
+    ],
+    lists:foreach(fun({Name, Eval}) ->
+        lager:info("===== simple: ~s =====", [Name]),
+        Eval()
+    end, Tests).
+
+big_circle() ->
+    State = big_circle_setup(),
```
mrallen1: This ought to be called `Nodes` since the state variable is really a list of Nodes.

tests/rt_cascading.erl
```diff
@@ -159,33 +198,9 @@ big_circle_test_() ->
...
-        rt:clean_cluster(Nodes)
-    end,
-    fun(Nodes) -> [
+    Tests = [
 
-        {"circle it", timeout, timeout(65), fun() ->
+        {"circle it", fun() ->
             [One | _] = Nodes,
             C = rt:pbc(One),
             Bin = <<"goober">>,
```
mrallen1: Another use of `<<objects>>` that ought to be `?bucket`

tests/rt_cascading.erl
```diff
@@ -1084,6 +1102,50 @@ wait_exit(Pids, Timeout) ->
...
+get_node(Name, NamesAndNodes) ->
+    [Node | _] = proplists:get_value(Name, NamesAndNodes),
+    Node.
+
+get_port(Name, NamesAndNodes) ->
+    get_cluster_mgr_port(get_node(Name, NamesAndNodes)).
+
 conf() ->
     [{lager, [
         {handlers, [
```
mrallen1: At line 1206, `WaitTime` ought to be called `SleepTime` and carried through.

At line 1220, where we start checking return values from `exists/3` we should use a new function name like `wait_until_obj_exists/6`.

tests/rt_cascading.erl
```diff
@@ -199,7 +214,7 @@ big_circle_test_() ->
             end || Node <- Nodes]
         end},
 
-        {"2 way repl, and circle it", timeout, timeout(65), fun() ->
+        {"2 way repl, and circle it", fun() ->
             ConnectTo = ["6", "1", "2", "3", "4", "5"],
             Connect = fun({Node, ConnectToName}) ->
                 Nth = list_to_integer(ConnectToName),
```
mrallen1: This fun should be renamed to ConnectNodes and the lists:map should be lists:foreach

tests/rt_cascading.erl
```diff
@@ -299,7 +321,7 @@ circle_test_() ->
             ?assertEqual(undefined, proplists:get_value(expect_seq, SinkData))
```
mrallen1: We should probably use `rt:wait_until` here instead of blindly hoping that this condition exists after 1 second.

tests/rt_cascading.erl
```diff
@@ -311,12 +333,41 @@ circle_test_() ->
             [SinkData] = proplists:get_value(sinks, Status, [[]]),
             ?assertEqual(2, proplists:get_value(expect_seq, SinkData))
```
mrallen1: Same issue here - we should use `rt:wait_until`

tests/rt_cascading.erl
```diff
@@ -385,29 +444,9 @@ diamond_test_() ->
...
-        rt:clean_cluster(Nodes)
-    end,
-    fun(Nodes) -> [
+    Tests = [
 
-        {"unfortunate double write", timeout, timeout(135), fun() ->
+        {"unfortunate double write", fun() ->
             [Top, MidLeft, MidRight, Bottom] = Nodes,
             Client = rt:pbc(Top),
             Bin = <<"start_at_top">>,
```
mrallen1: Wow, sleep for 100 seconds??! Explore rt:wait_until

tests/rt_cascading.erl
```diff
@@ -535,25 +582,66 @@ circle_and_spurs_test_() ->
...
-        ok ->
-            pass;
-        error ->
-            % at the time this is written, the return value isn't acutally
-            % checked, the only way to fail is to crash the process.
-            % i leave the fail here in hopes a future version will actually
-            % do what the documentation says.
-            exit(error),
-            fail
+    case rt_config:config_or_os_env(run_rt_cascading_1_3_tests, false) of
```
mrallen1: Are these tests still needed ?? Or do they need to be updated for "new" Riak versions like 2.0 -> 2.1?  Seems like these are specifically testing for 1.3 -> 1.4 changes.

tests/rt_cascading.erl
mrallen1: At 1259 looking at all of the calls into repl_util, we need to refactor `repl_util:wait_for_reads` to just return the value from `rt:wait_until` - if it returns `ok` then we don't need to rerun the reads again, if it throws an assertion failure the test should fail anyway, so the code in repl_util at line 203 and following is redundant.

This needs to be examined across other repl tests to see if they're affected.

PR #936, opened 2015-11-23
Add tests for utf8 and latin1 in the table definition

andytill:
Neither is currently supported because of the way JSON is encoded, when it is supported these tests should be modified to assert on correct table creation.


tests/ts_unicode_2.erl
```diff
@@ -0,0 +1,45 @@
...
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% Test for creating a utf8 table name. This should fail
+%% because of mochijson decoding.
+
+-module(ts_unicode_2).
```
javajolt: Andy -

Would it be useful to expand the existing `ts_unicode` test to encompass your changes?

andytill: I prefer one assert per test, is there any benefit?

javajolt: Just less bookkeeping.

How would you feel about naming your test something like `ts_unicode_create_table_not_allowed`? I'm always trying to get more explicit test names.

tests/ts_latin1.erl
```diff
@@ -0,0 +1,45 @@
...
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% Test for creating a latin1 table name. This should fail
+%% because of mochijson decoding.
+
+-module(ts_latin1).
```
javajolt: Same for this test. Perhaps call it `ts_latin1_create_table_not_allowed`?

javajolt:
```
Test Results:
ts_unicode_2-eleveldb: pass
ts_latin1-eleveldb   : pass
0 Tests Failed
2 Tests Passed
```

javajolt:
:+1:

PR #937, opened 2015-11-23
Fix race in repl_rt_heartbeat due to short timeout

nickelization:
One particular timeout in the repl_rt_heartbeat test was slightly too
short, which could cause us to occasionally hit a false positive on this
test if various timings lined up just right. This PR bumps up the
timeout, which should prevent this from happening again.

I would really like to do a proper fix for this, which would use
intercepts or something to confirm that the actual timeout is being hit
in the code...but we don't really have time for that, and a half fix is
better than no fix I suppose.


javajolt:
Agreed that this is more of a band-aid, but I bet it'll buy us a bit more time.

javajolt:
:+1: 2be9c2f

PR #938, opened 2015-11-23
Update the error message for ts_A_select_compare_two_fields_not_allowed

javajolt:
The previous error message was temporary.

### DIFF
--- a/tests/ts_A_select_compare_two_fields_not_allowed.erl
+++ b/tests/ts_A_select_compare_two_fields_not_allowed.erl
@@ -39,7 +39,6 @@ confirm() ->
         "AND myfamily = 'fa2mily1' "
         "AND myseries ='seriesX' "
         "AND weather = myseries",
-    Expected = "Expect that fields cannot be compared",
-    Got = ts_util:ts_query(ts_util:cluster_and_connect(single), normal, DDL, Data, Qry),
-    ?assertEqual(Expected, Got),
+    {error, Got} = ts_util:ts_query(ts_util:cluster_and_connect(single), normal, DDL, Data, Qry),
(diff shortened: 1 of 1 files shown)


tests/ts_A_select_compare_two_fields_not_allowed.erl
```diff
@@ -39,7 +39,6 @@ confirm() ->
         "AND myfamily = 'fa2mily1' "
         "AND myseries ='seriesX' "
         "AND weather = myseries",
-    Expected = "Expect that fields cannot be compared",
-    Got = ts_util:ts_query(ts_util:cluster_and_connect(single), normal, DDL, Data, Qry),
-    ?assertEqual(Expected, Got),
+    {error, Got} = ts_util:ts_query(ts_util:cluster_and_connect(single), normal, DDL, Data, Qry),
+    ?assertNotEqual(0, string:str(binary_to_list(Got), "Comparing or otherwise operating on two fields is not supported")),
```
macintux: I'd rather assert Got as equal to the current error message, even with the undesirable but probably going to linger indefinitely "Message decode" prefix.

javajolt: Really? I just refactored the test to avoid that case. Why would you like to make it more brittle when we are (hopefully) going to change the code/message?

macintux: I suppose that since you're matching an error tuple, the test has some value, but it seems odd to fail on the error message you actually want to see.

The test could fail for any error scenario, which seems less desirable. I'd rather fix the error message when (**if**) we fix the backend.

macintux: Ok, never mind, move along, nothing to see here except a facepalm.

PR #939, opened 2015-11-24
Log the start of each test in repl_aae_fullsync

nickelization:
This makes it much easier to tell which code is being executed at a given point in the log, since many of the tests print identical log messages.


JeetKunDoug:
:+1:

PR #942, opened 2015-11-25
Update tests for different default bucket properties

nickelization:
In Riak 2.1.2 there's been a small, subtle change made in the way that default bucket properties work. To ensure all our tests keep working the same way, we need to explicitly specify settings for allow_mult and dvv_enabled whenever we have a custom value set for riak_core.default_bucket_props. Most of these changes may not be strictly necessary, but we do at least need the change to repl_aae_fullsync, and in general it's better to be explicit about what we expect.

For more background, see https://github.com/basho/riak_core/pull/765

This PR also includes a couple of indentation fixes, and a small, unrelated change that adds some extra logging to wait_until_no_pending_changes.

Doug and I ran through each of the modified tests as well, just to verify that nothing was broken by these changes.


JeetKunDoug:
:+1:

PR #943, opened 2015-11-25
new ts_A_create_table_* modules for more comprehensive ddl validation

hmmr:
Covers RTS-616, RTS-620, RTS-623 (items D, H, J in RTS-610).

Depends on https://github.com/basho/riak_kv/pull/1279 and https://github.com/basho/riak_ql/pull/50. Supersedes https://github.com/basho/riak_test/pull/941.


tests/ts_A_create_table_already_created.erl
```diff
@@ -1,4 +1,3 @@
-%% -*- Mode: Erlang -*-
 %% -------------------------------------------------------------------
```
javajolt: What do you have against this emacs or vim encoding?

hmmr: But.. emacs would only need this if you had your sources ending in something other than .erl. And, some odd ones excluding, this line is not used anywhere else in riak.

tests/ts_A_create_table_short_key.erl
```diff
@@ -32,7 +31,7 @@
 confirm() ->
     ClusterType = single,
     DDL = ts_util:get_ddl(shortkey_fail),
-    Expected = {ok,"Error creating bucket type GeoCheckin:\nPrimary key is too short\n"},
+    Expected = {ok,"Error validating table definition for bucket type GeoCheckin:\nPrimary key must consist of exactly 3 fields (has 2)\n"},
     Got = ts_util:create_bucket_type(ts_util:build_cluster(ClusterType), DDL),
```
javajolt: What if is only has 1? Maybe we should skip the (has 2) part?

tests/ts_A_create_table_already_created.erl
javajolt: OK. Erik added it to his tests so I assumed he needed it for something.

tests/ts_A_create_table_not_null_pk_fields.erl
```diff
@@ -0,0 +1,37 @@
...
+
+-include_lib("eunit/include/eunit.hrl").
+
+-export([
+         confirm/0
+        ]).
+
+confirm() ->
+    ClusterType = single,
+    DDL = ts_util:get_ddl(not_null_primary_key_field_fail),
```
andytill: I would much prefer if this DDL was in the test and not ts_util, the data in a test is important to see to understand the test.

hmmr: It would make sense to do this globally, for all ts_* modules. @javajolt if you are ok with this, I undertake to do this (in this PR).

andytill: That would be great. If some table definitions are shared by many tests the readability trade off probably isn't worth it and it is still better to share them. Another problem caused by riak_test not having test suites like common test.

javajolt: @hmmr Not sure who designed this originally (Gordon?), but I'm quite happy with the the SQL being with the tests as @andytill suggests.  As he also said, the advantage to having (some) of the basic SQL/DDL in `ts_utils` is that we don't have lots of duplication and updates can be made in one place.  So if it's used in multiple places, let's keep it  in `ts_utils` for now, but move unique SQL/DDL to the respective tests.

hmmr:
Disbanded single-use `get_ddl/1` and moved the DDL strings directly to where they are used in ts_A_create_table_*_.

Also, reindent, shorten lines, de-tab.

tests/ts_A_create_table_dup_primary_key.erl
```diff
@@ -0,0 +1,46 @@
...
+-module(ts_A_create_table_dup_primary_key).
+
+-behavior(riak_test).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-export([confirm/0]).
+
+confirm() ->
+    DDL =
```
javajolt: Seems like this is `ts_utils:get_ddl()`

javajolt:
:+1: 16214fa

PR #944, opened 2015-11-30
Try to activate the bucket type 4 times

javajolt:
It's a bit ugly because the return result from `rt:admin` is always the tuple `{ok, Msg}` so I need to look for the relevant parts of that message to determine activation success.  If we timeout, then that same result needs to be returned to the caller.


macintux:
:+1:

PR #946, opened 2015-12-01
Update proxy_overload_recovery to handle indeterminism more gracefully.

JeetKunDoug:
- update `prepare` so that it waits for the vnode PID to change as it's
  possible that monitors haven't fired by the time we ask for a new one
- Update `drain` to require 3 consecutive 0 counts from a group of PIDs
  rather than draining individual processes or only trying once. The
  number 3 was determined empirically - read the comment above `drain`
  if it continues to cause issues.
- Rewrite `resume` to make sure things are drained properly before
  continuing.


jonmeredith:
+1 9551318edd3a94ea0a9a806bd8f69d1573fe3940

PR #948, opened 2015-12-01
Log PIDs by default for console output

nickelization:
We've run into a number of issues over time where errant processes hang
around and stomp on other tests, and without PIDs in the logging it can
be difficult or impossible to conclusively identify and debug these
kinds of problems.


JeetKunDoug:
This should resolve RIAK-2306

javajolt:
Works as advertised
:+1:

PR #951, opened 2015-12-02
RTS-531: Use some ts_util functions in ts_basic to avoid code duplication

javajolt:
In particular we want to leverage the retries in bucket-type activation


macintux:
:+1:

PR #952, opened 2015-12-03
Update test for new error message

macintux:


javajolt:
Even better suggestion: Could we pattern match on the error string? Or do we want to keep it static?

macintux:
I thought about that, but it seems unlikely to change at this point.

javajolt:
:+1: f313841

PR #953, opened 2015-12-06
include a stream_list_keys in ts_basic

hmmr:
RTS-697. Requires https://github.com/basho/riak_kv/pull/1291.

With changes in https://github.com/basho/riak_kv/pull/1291, ts_basic will have `riakc_ts:list_keys/3` scrapped and replaced by a wrapper around `riakc_ts:stream_list_keys/3`.

The wrapper works by repeatedly calling and collecting keys from the streaming version of list_keys until the required, known, number of keys is received. This becomes necessary because the streaming version, working asynchronously, takes some time to return the keys intended to be in the bucket.

### DIFF
--- a/bin/rtdev-setup-releases.sh
+++ b/bin/rtdev-setup-releases.sh
@@ -42,6 +42,10 @@ git init
 git config user.name "Riak Test"
 git config user.email "dev@basho.com"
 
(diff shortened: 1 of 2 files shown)


javajolt:
:+1:

PR #954, opened 2015-12-07
Test term_to_binary

macintux:
Copied `ts_basic` and added a toggle for native Erlang encoding.

See basho/riak_pb#171 (RIAK-1299) for more details


javajolt:
How about we just add this as another test step within `ts_basic`? Then we don't have the copy pasta

macintux:
That's reasonable. I'll rework this later.

javajolt:
:+1:

PR #955, opened 2015-12-08
Pause to allow delete_mode delay to function

macintux:
Key lists have historically included tombstones. In TS they have not because Andrei's code had to open each object to determine the key; with @jonmeredith's forthcoming changes, we can now reliably generate the key tuple without opening the object.

Because of this, we'll start seeing `ts_basic` failures because the list keys function still sees the deleted object because the tombstone has yet to be reaped. Introduce a small delay to allow reaping.


JeetKunDoug:
:+1: Saw failure w/o fixes with correct riak_kv branch, and passes several times with fixes

PR #959, opened 2015-12-15
check data overwrite

hmmr:
Check if the data can be overwritten.


javajolt:
@hmmr Can we work on breaking this test up?  Or at least create a second test with the sole purpose of overwriting data?

hmmr:
I will create a separate test module, just for that (but, having it also included in ts_mighty_basic wouldn't hurt).

javajolt:
Let's get it in so we at least have test coverage

javajolt:
:+1:

PR #962, opened 2015-12-18
Add new Time Series 1.1 Tests

javajolt:
These primarily cover aggregation, but also include arithmetic and the schema DESCRIBE function.


tests/ts_describe_table.erl
```diff
@@ -0,0 +1,40 @@
...
+%% Test basic aggregation functions
+
+confirm() ->
+    TestType = normal,
+    DDL = ts_util:get_ddl(),
+    Data = ts_util:get_valid_select_data(),
+    Qry = "DESCRIBE GeoCheckin",
+    Got = ts_util:ts_query(
+        ts_util:cluster_and_connect(single), TestType, DDL, Data, Qry),
+    Expected = some_table_description,
```
hmmr: This is what to expect from the query:
```
Expected =
        {[<<"Column">>,<<"Type">>,<<"Is Null">>,<<"Primary Key">>, <<"Local Key">>],
         [{<<"myfamily">>,<<"varchar">>,   false,  1,  1},
          {<<"myseries">>,<<"varchar">>,   false,  2,  2},
          {<<"time">>,    <<"timestamp">>, false,  3,  3},
          {<<"weather">>, <<"varchar">>,   false, [], []},
          {<<"temperature">>,<<"double">>, true,  [], []}]},
```

tests/ts_aggregation_math.erl
```diff
@@ -0,0 +1,65 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(ts_aggregation_math).
```
andytill: It is written maths :)

javajolt: Silly 'muricans and our inability to spell

javajolt:
@hmmr How does it look now? It passes, BTW. The other tests will fail until Gordon is ready, but would like to get them in place so we can tweak to pass.

PR #964, opened 2015-12-19
new module to check fields are defined in order they appear in keys

hmmr:
RTS-786.

Depends on https://github.com/basho/riak_ql/pull/71.


hmmr:
This issue is covered by a eunit test, which means it is bound to pass. Closing.

PR #967, opened 2015-12-30
Row validation test

andytill:
A test asserting that all rows in a batch are validated.  The error message is not fully specified to make the tests less brittle. This is covered in the eunit tests for building the error message here: https://github.com/basho/riak_kv/commit/3f7555a69ccddc6ac2d266042f85974c5826cc0b

### DIFF
--- a/tests/ts_A_put_invalid_data.erl
+++ b/tests/ts_A_put_invalid_data.erl
@@ -39,22 +39,23 @@ confirm() ->
     LongObj = ts_util:get_long_obj(),
     Bucket = ts_util:get_default_bucket(),
     {_Cluster, Conn} = ClusterConn = ts_util:cluster_and_connect(single),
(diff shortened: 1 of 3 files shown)


andytill:
I changed the test specifically not to assert on the error messages byte for byte. This is easy to do but has caused lots of test failures when error messages are changed and improved upon.

The eunits in the corresponding riak_kv PR test the error messages are built correctly.

hmmr:
I strongly agree with @andytill on matching on return codes only. Just, doing this in an isolated PR will uncover a mighty lot of similar cases in ts_*. Still, +1 here.

andytill:
I had committed the wrong file -__- which has now been blasted because I thought I already committed it.  I imagine my comments made no sense at all without the proper file.  Suggest we go with this as-is.

PR #971, opened 2016-01-08
Test that the COUNT function ignores NULL

andytill:
If a single column is used as the input to COUNT then any NULL values are not included in the total.

Testing https://github.com/basho/riak_ql/pull/87.


tests/ts_aggregation_simple.erl
```diff
@@ -77,7 +84,9 @@ verify_aggregation(ClusterType) ->
        <<"COUNT(temperature)">>,
        <<"COUNT(precipitation)">>
       ],
-      [{Count, Count, Count}]},
+      [{count_non_nulls(?PRESSURE_COL_INDEX, Data),
```
javajolt: @andytill I thought this test had previously passed.  It is simply because not an entire row was NULL?

andytill: It is because the `COUNT` is called with one column so `NULL` is not counted in the total e.g. `COUNT(precipitation)`.  The counting on the test side now checks for `NULL` in the data written as well.

PR #972, opened 2016-01-08
Rename STDEV to STDDEV

gordonguthrie:


javajolt:
I wondered about that. I thought STDDEV was a bit more...er... standard.

javajolt:
:+1:

PR #973, opened 2016-01-11
Switch from ts_A, B, C to ts_simple, cluster, degraded

javajolt:
Change the naming convention to be a bit more human-readable.  Tests were verified that `ts_cluster` tests now build a cluster and `ts_simple` tests are for a single node.


gordonguthrie:
+1 f2f15c8

PR #986, opened 2016-02-02
Add new test verify_stats_removal (RIAK-2312)

JeetKunDoug:
When a vnode starts up, it creates several vnode-specific stats. When that vnode shuts down (generally due to handoff) those stats were not correctly removed. This test confirms that in the happy-path shutdown case those stats are, in fact, removed. Future work in the next release will handle the vnode crashing as well.

This test verifies the fix in basho/riak_kv#1282

### DIFF
--- a/tests/verify_stats_removal.erl
+++ b/tests/verify_stats_removal.erl
@@ -36,11 +36,10 @@
 confirm() ->
     Nodes = rt:deploy_nodes(2),
     [Node1, Node2] = Nodes,
-    %% Need to write some data to pump the stats system - seems at least 5 values before it starts counting
-    %% stolen from verify_riak_stats
+    %% Need to write some data to pump the stats system
     C = rt:httpc(Node1),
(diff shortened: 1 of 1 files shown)


tests/verify_stats_removal.erl
```diff
@@ -0,0 +1,77 @@
...
+    pass.
+
+get_stats_count_for_non_running_vnodes(Node1) ->
+    spawn_link(Node1, verify_stats_removal, get_stats_remote, [self()]),
+    receive
+        {stats, NumStats} ->
+            NumStats;
+        _Other ->
+            throw("Failed to retrieve count of stats from vnode.")
+    end.
```
jonmeredith: May be worth a longish timeout (60s) rather than hang r_t runs.

JeetKunDoug:
Added timeout - test should be ready to go. Confirms that stats are leaking on non-fixed code, not leaking on fixed.

mrallen1:
:+1: 6e63512

mrallen1:
Passes against riak_kv 2.0 08edc1
Fails against riak_kv 2.0 290b35

PR #987, opened 2016-02-02
RTS-876: Clean up ts_simple_aggregation_fail test failures

javajolt:
Needs these PRs: https://github.com/basho/riak_ql/pull/102 and https://github.com/basho/riak_kv/pull/1339


javajolt:
@hmmr All of the error codes seem to be the same.  Andy just wanted to add a regex comparison.  Maybe that would be a better solution for the string comparison?

hmmr:
Regexes will certainly also do.

javajolt:
@hmmr Now with a regex instead.  See if that works for you.

PR #989, opened 2016-02-03
RTS-890: Change expected error message for ts_simple_select_unexpected_token_not_allowed

javajolt:
Add the error code `1020` to the expected error message for unexpected tokens.

### DIFF
--- a/tests/ts_simple_select_unexpected_token_not_allowed.erl
+++ b/tests/ts_simple_select_unexpected_token_not_allowed.erl
@@ -33,7 +33,7 @@ confirm() ->
     Qry =
         "selectah * from GeoCheckin "
         "Where time > 1 and time < 10",
-    {error, Got} = ts_util:ts_query(
+    Got = ts_util:ts_query(
                      ts_util:cluster_and_connect(single), TestType, DDL, Data, Qry),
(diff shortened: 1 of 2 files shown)


tests/ts_simple_select_unexpected_token_not_allowed.erl
```diff
@@ -33,7 +33,7 @@ confirm() ->
     Qry =
         "selectah * from GeoCheckin "
         "Where time > 1 and time < 10",
-    {error, Got} = ts_util:ts_query(
+    {error, {1020, Got}} = ts_util:ts_query(
                      ts_util:cluster_and_connect(single), TestType, DDL, Data, Qry),
     ?assertNotEqual(0, string:str(binary_to_list(Got), "Unexpected token")),
```
andytill: This is getting a little hard to understand what is being asserted with `Got` being part of what is gotten and the string manipulation. How about a custom assertion, something like.
```erlang
assert_error_like(
    {error, {1020, <<".*Unknown Token.*">>}},
    do_the_thing()
).
```
The assertion understand the common error formats received from Time Series to match against, and the binary is a regex. This would make it really flexible, since we generally need less brittle tests for error messages.

andytill:
+1 d6a30c9

PR #990, opened 2016-02-04
Return an error when deleting a key that does not exist RTS-894

andytill:
Currently `ok` is returned when a single key delete is issued against a key that does not exist.

With this change `{error, {1021, <<"notfound">>}}` is returned instead. 1021 is a new error code for the case where the key does not exist.


javajolt:
@andytill Does this depend upon https://github.com/basho/riak_kv/pull/1341?

javajolt:
:+1:

PR #991, opened 2016-02-08
TS Flexible key tests

andytill:


tests/ts_cluster_keys.erl
```diff
@@ -0,0 +1,169 @@
+-module(ts_cluster_keys).
```
javajolt: Sorry to be nit picky here, @andytill, but how about `ts_cluster_flexible_keys`? Just `keys` seems a wee bit vague.

javajolt:
Guessing that https://github.com/basho/riak_kv/pull/1333 is required for this PR, too.

tests/ts_cluster_keys.erl
andytill: The key changes seem flexible now but in the feature we will think of them just as keys.

andytill:
Closing, going to open a new PR to 1.3

PR #992, opened 2016-02-09
Single key op tests (deletes)

andytill:
Currently only contains single key deletes. Shows failures where the key fields are not consecutive on the table definition.

### DIFF
--- a/tests/ts_simple_single_key_ops.erl
+++ b/tests/ts_simple_single_key_ops.erl
@@ -1,0 +1,87 @@
+-module(ts_simple_single_key_ops).
+-behavior(riak_test).
+-compile([export_all]).
+-export([confirm/0]).
+-include_lib("eunit/include/eunit.hrl").
+
+confirm() ->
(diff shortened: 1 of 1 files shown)


tests/ts_simple_single_key_ops.erl
```diff
@@ -0,0 +1,87 @@
...
+	?assertEqual(
+		ok,
+		riakc_ts:delete(Pid, <<"table1">>, [1,2,4], [])
+	).
+
+%%%
+%%% TABLE 2
+%%%
+
+% table_def_2() ->
```
javajolt: Does this test not pass?  Just wonder since it is commented out.

andytill: It doesn't pass, it is for flexible keys.

javajolt:
Looks like https://github.com/basho/riak_kv/pull/1345 is required

javajolt:
:+1:

PR #1003, opened 2016-02-21
Change assert_float to use relative error

javajolt:
#Tweak how `ts_util:assert_float` determines the difference between too floats.  Previously two values were calculated by taking the difference over the average of the numbers.  Now use relative error which uses the absolute value of the difference over the expected value. https://en.wikipedia.org/wiki/Approximation_error

Also the aggregation function STDDEV had been tested to be equal to STDDEV_POP which is incorrect. It's actually the same as STDDEV_SAMP and the above change helps to detect the difference.

### DIFF
--- a/tests/ts_util.erl
+++ b/tests/ts_util.erl
@@ -470,7 +470,7 @@ get_optional(N, X) ->
     end.
 
 
--define(DELTA, 1.0e-10).
+-define(EPSILON, 1.0e-10).
 
 assert_float(String, {Cols, [ValsA]} = Exp, {Cols, [ValsB]} = Got) ->
     case assertf2(tuple_to_list(ValsA), tuple_to_list(ValsB)) of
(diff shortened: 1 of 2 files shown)


tests/ts_util.erl
```diff
@@ -486,9 +486,9 @@ assert_float(String, Exp, Got) -> assert(String, Exp, Got).
 
 assertf2([], []) -> pass;
 assertf2([H1 | T1], [H2 | T2]) ->
-    Diff = H1 - H2,
-    Av = (H1 + H2)/2,
-    if Diff/Av > ?DELTA -> fail;
+    Diff = abs(H1 - H2),
+    Delta = Diff/H2,
```
macintux: This looks like it won't work properly for negative floats; `H2` should also be an absolute value (or just absolute value `Delta`).

PR #1004, opened 2016-02-29
Add basic test for riak_kv#1356

russelldb:
Start riak, add data, stop riak, read data. With known bad level
config.


martinsumner:
Merged instead via https://github.com/basho/riak_test/pull/1330

PR #1007, opened 2016-03-01
Add tests for readrepair

bsparrow435:
Tests to verify that object limits are ignored for read repair. This
requires an intercept to spoof all puts as read repair since there’s no
external API for forcing read repair. Intercept used to replace both
put/6 and coord_put/6 and simply adds the `rr` property to the put
options.


mrallen1:
This PR targeted the wrong branch and throws undefs at `rt_intercept:clean/2`

PR #1009, opened 2016-03-02
Add tests for read repair object limits

bsparrow435:
Tests to verify that object limits are ignored for read repair. This
requires an intercept to spoof all puts as read repair since there’s no
external API for forcing read repair. Intercept used to replace both
put/6 and coord_put/6 and simply adds the rr property to the put
options.

### DIFF
--- a/tests/verify_object_limits.erl
+++ b/tests/verify_object_limits.erl
@@ -36,6 +36,7 @@
 confirm() ->
     [Node1] = rt:build_cluster(1, [{riak_kv, [
                         {ring_creation_size, 8},
+                        {anti_entropy, {off,[]}},
                         {max_object_size, ?MAX_SIZE},
                         {warn_object_size, ?WARN_SIZE},
                         {max_siblings, ?MAX_SIBLINGS},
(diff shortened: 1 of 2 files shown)


mrallen1:
:+1: d66bf75

mrallen1:
**N.B.** Requires basho/riak_kv#1363 to pass

tests/verify_object_limits.erl
```diff
@@ -128,3 +132,49 @@ verify_sibling_limits(C, Node1) ->
     lager:info("Result when too many siblings : ~p", [Res]),
     ?assertMatch({error,_},  Res),
     ok.
+
+verify_readrepair_ignore_max_size(C, Node1) ->
+    % Add intercept to treat all vnode puts as readrepairs
+    Intercept = {riak_kv_vnode, [{{put, 6}, put_as_readrepair},{{coord_put,6}, coord_put_as_readrepair}]},
+    ok = rt_intercept:add(Node1, Intercept),
+    timer:sleep(100),
```
nickelization: @bsparrow435 is this timer:sleep call necessary? I don't think you're supposed to have to sleep after adding an intercept (or if we are, I would hope there's some way of doing it in a deterministic way, rather than just hoping 100 ms is long enough to wait). I tried taking this out - along with the other timer:sleep call you added down below on line 166 - and the test seems to run fine without them.

bsparrow435: Yes you're right sorry. This was added as a debug step when the test was failing for me for an unrelated reason. I saw some sleeps in intercept adds in other tests and was trying it out. I agree, it's not necessary and can be removed.

mrallen1:
:+1: 1d6ab06

PR #1011, opened 2016-03-04
Test TS batching

macintux:
Test the integrity of batch uploads. Supports basho/riak_kv#1366


hmmr:
I would suggest the module to (a) set `timeseries_max_batch_size` explicitly, and (b) use intercepts (like so: https://github.com/basho/riak_test/pull/914/files#diff-1273dcf245502ec493e64d8c4009ffe1R44) to verify the `riak_kv_eleveldb_backend:batch_put` is invoked the expected number of times. Otherwise, the test amounts to a simple retrieval of the data you have just inserted, which will surely pass (or should it fail, it will not obviously fail for the batching put path being broken).

macintux:
I shared that concern, although I think you overestimate the likelihood of success if something, anywhere, is wrong with the code. :-)

I’ll try to make time to make this better.

-John

macintux:
Ok, @hmmr, finally batching is tested properly.

tests/ts_simple_batch.erl
```diff
@@ -36,15 +36,41 @@
...
+    rt_intercept:add(Node, {riak_kv_eleveldb_backend,
+                            [{{batch_put, 4}, batch_put}]}),
+
+    %% Buried in the bowels of the code path behind ts_util:ts_put/4
+    %% is a calculation that n_val is the same as the cluster size. I
+    %% want a single node cluster for this test, but n_val of 4, so
+    %% I'll duplicate the path here
+    Bucket = ts_util:get_default_bucket(),
+    {ok, _} = ts_util:create_bucket_type([Node], DDL, Bucket, 4),
+    ts_util:activate_bucket_type([Node], Bucket),
```
hmmr: I would prefer `riak_ts:query(rt:pbc(Node), DDL ++ " WITH (nval=4)")`. But it's not something that needs 'fixing' here.

PR #1016, opened 2016-03-18
test setting TS table bucket type's props with WITH clause

hmmr:
RTS-865

Tests for WITH-clause properties implemented in https://github.com/basho/riak_kv/pull/1369.


tests/ts_cluster_create_table_via_sql.erl
```diff
@@ -26,30 +26,65 @@
...
+-define(CUSTOM_MVAL, <<"bo o''oo">>).
+
 confirm() ->
     [Node1, Node2 | _] = _Cluster = ts_util:build_cluster(multiple),
     C1 = rt:pbc(Node1),
     C2 = rt:pbc(Node2),
 
-    DDL = ts_util:get_ddl(),
-    Table = ts_util:get_default_bucket(),
+    CommonDDL =
```
macintux: Instead of copying the DDL here to allow for table name customization, extend `ts_util`. This is a good feature for parallelization of tests. `get_ddl(Size, TableName)` e.g.

tests/ts_cluster_create_table_via_sql_SUITE.erl
```diff
@@ -0,0 +1,140 @@
...
+    Got1 = riakc_ts:query(C, GoodDDL),
+    ?assertEqual({[],[]}, Got1),
+    io:format("Created table via query: ~p\n", [Got1]),
+    pass.
+
+re_create_fail_test(Ctx) ->
+    C = client_pid(Ctx),
+    Got = riakc_ts:query(C, ddl_common()),
+    ?assertMatch({error, {1014, _}}, Got),
+    io:format("Not created same table via query: ~p\n", [Got]),
```
andytill: This logging is confusing while the test is running. It is not clear whether this is logging a good thing that happened or a bad thing without looking at the code. We shouldn't log anything extra when the test passes.

tests/ts_cluster_create_table_via_sql_SUITE.erl
```diff
@@ -0,0 +1,140 @@
...
+        end,
+    ?assertEqual(ExpectedPL, GetBucketPropsF(Node1)),
+    ?assertEqual(ExpectedPL, GetBucketPropsF(Node2)),
+    io:format("Verified these properties are set: ~p\n", [ExpectedPL]),
+    pass.
+
+
+client_pid(Ctx) ->
+    client_pid(1, Ctx).
+client_pid(N, Ctx) ->
```
andytill: `client_pid/2` is not used.

tests/ts_cluster_create_table_via_sql_SUITE.erl
```diff
@@ -0,0 +1,140 @@
...
+
+table_described() ->
+    {[<<"Column">>,<<"Type">>,<<"Is Null">>,<<"Primary Key">>, <<"Local Key">>],
+     [{<<"myfamily">>,   <<"varchar">>,   false,  1,  1},
+      {<<"myseries">>,   <<"varchar">>,   false,  2,  2},
+      {<<"time">>,       <<"timestamp">>, false,  3,  3},
+      {<<"weather">>,    <<"varchar">>,   false, [], []},
+      {<<"temperature">>,<<"double">>,    true,  [], []}]}.
+
+unenquote_varchars(PP) ->
```
andytill: This seems to only be used to modify `?CUSTOM_MVAL` because the value has `''` in it. Can the value just have `'` in it?

hmmr: Of course it can, but there is an educational angle to it. This function is here to demonstrate the (non-intuitive) way single quotes are "escaped", and also to suggest that string values are parsed as varchars.

andytill: Ah so `bo o''oo` goes in but the value is actually `bo o'oo`. It is educational!

I wouldn't have guessed that this is what is happening in the test though, only one thing should be tested per test. Encoding should have it's own test. The other test shouldn't use a property with `''` in it so if the encoding is broken during dev it is obvious that the problem is isolated to just that.

javajolt:
```
Test Results:
ts_cluster_create_table_via_sql_SUITE-riak_kv_bitcask_backend,riak_kv_eleveldb_backend: pass
0 Tests Failed
1 Tests Passed
```
:+1:

hmmr:
There appears to be an issue with setting string properties that have `'` at the end or beginning of the string value. It's probably a parser issue (and so belongs in riak_ql), but it's still not exactly clear what and where needs to be fixed. So please hold off the merge.

hmmr:
Now the test string includes a final `'`, which will test how the lexer copes with this case, and now we depend on https://github.com/basho/riak_ql/pull/116.

javajolt:
:+1: f4824e1

PR #1017, opened 2016-03-19
update tests for new default bucket props 2.0

JeetKunDoug:
This is a backport of #942 from the riak/2.1 branch, as we've now included this fix in Riak 2.0 as well (well, was always in the 2.0 branch, but we hadn't released it yet). Will resolve RIAK-2377.


nickelization:
+1 eb4949b Looks like a straightforward backport; the whole patch looks the same, other than the two write-once tests we modified in 2.1, which don't exist in 2.0.

PR #1018, opened 2016-03-21
RTS-949: Add @seemaj's riak_shell script as a test

javajolt:
Add a `riak_shell` regression log as a basic functionality test.  Also fix interaction with `riak_shell` which was changed in https://github.com/basho/riak_shell/pull/23. We are consuming the command `cmd_error` code, but are not using it in tests yet.

Requires https://github.com/basho/riak_shell/pull/27


hmmr:
The log gets replayed and the test passes, so there's my +1.

However, going outside the scope of this PR, it appears a new instance of the riak_shell app is started on the test node (in addition to those started on the dev? nodes). As a standalone application, riak_shell can be started anywhere, but I wonder if you could just use the one running on dev1? This way, you would probably not need to have those kludgy `riak_shell:*_TEST` functions.

hmmr:
+1 e0cccaf

javajolt:
@hmmr Gordon set up the testing infrastructure for `riak_shell`.  Although the binary lives in dev1, it isn't actually running.  I think the complication is how the shell interacts with the command prompt.  He had to set up separate loop to inject input into the shell without actually waiting for input from the TTY.  I agree it's complicated, but it gets us there in the end.

PR #1019, opened 2016-03-21
Update basic_command_line for new CLI messages

nickelization:
In the latest 2.0 code, it now prints "node did not respond to ping"
instead of "node is not running" when using a ping test to try and
check the node status.

### DIFF
--- a/tests/basic_command_line.erl
+++ b/tests/basic_command_line.erl
@@ -24,6 +24,8 @@
 -compile(export_all).
 -export([confirm/0]).
 
+-define(PING_FAILURE_OUTPUT, "Node did not respond to ping!").
+
 confirm() ->
 
     %% Deploy a node to test against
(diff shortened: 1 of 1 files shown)


mrallen1:
Wondering if it might be worthwhile to use a macro for this string in case it changes again.

nickelization:
Thanks, good call. I will update this PR soon with an additional commit.

nickelization:
@mrallen1 - PR updated with a new commit moving that string into a macro.

mrallen1:
:+1: 4dfcc74

javajolt:
@nickelization Is this change also in 2.1+?

nickelization:
@javajolt not yet. This change was introduced in node_package 3.0.0, and other versions of Riak are still using 2.x.x.

PR #1020, opened 2016-03-21
Handle node_package message change

tburghart:
Adds rt:str_mult/2, to check for any of multiple substrings.
Alters the basic_command_line test to accept the ping response from node_package 2.x or 3.x.


JeetKunDoug:
:+1: thanks for fixing this. Am in the process of cherry-picking the commits & pushing it to the riak/2.1 (and eventually riak/2.0) branches.

PR #1022, opened 2016-03-23
Merge the latest from develop to develop-2.2

javajolt:


nickelization:
:+1: Fast forward merge, so should be safe, and makes sense to me since develop-2.2 is the latest version-specific develop branch.

PR #1027, opened 2016-03-25
integrate common test with riak_test

gordonguthrie:


javajolt:
All the TS, riak_test tests pass. I'll try @andytill's flexible keys one now.

src/riak_test_escript.erl
```diff
@@ -429,3 +444,13 @@ so_kill_riak_maybe() ->
             io:format("Leaving Riak Up... "),
             rt:whats_up()
     end.
+
+is_common_test([]) -> 
+    false;
+is_common_test([{attribute, _, file, {File, _}} | T]) ->
+    case filename:basename(File) of
```
tburghart: It seems there must be a better way to test this, isn't there?

javajolt: Do you have a suggestion, @tburghart?

andytill: rebar requires all common tests end in `_SUITE.erl`.

javajolt: @andytill Is that true of rebar3, too?

andytill: No idea, I'm an erlang.mk guy ;)

rebar2 at least required the module name without the SUITE and then add it to the end to form the file name.

src/riak_test_escript.erl
```diff
@@ -429,3 +444,13 @@ so_kill_riak_maybe() ->
...
+        "ct.hrl" -> true;
```
hmmr: Will this work if the source has `-include_lib("common_test/include/ct.hrl").`?

javajolt:
I tried three of @andytill's tests from his CT PR:
- ts_cluster_keys_SUITE
- ts_cluster_quantum_boundaries_SUITE
- ts_updown_select_SUITE

The first two passed, but the third failed.  I suspect it need some other changes, to pass, however.

andytill:
@javajolt `ts_updown_select_SUITE.erl` was just a test suite for developing the common test integration, it is just a copy/pasta of one of the other basic tests. It should be removed.  Shall I do this here or does it need to go on another branch?

javajolt:
OK, if that is a red herring, then let's get this thing merged.

javajolt:
:+1:

PR #1030, opened 2016-03-30
Fix issues with yz_extractor test

nickelization:
This PR addresses some false positives we were occasionally seeing with the yz_extractor test. The issue had to do with some test code that counts the number of calls to an internal yokozuna function, get_map. It normally gets called once per read, so after doing 20 reads with n=3, the test expects to see that the function was called 60 times. Occasionally though, a YZ AAE exchange would occur during those 20 writes and trigger a repair, which in turn would trigger an extra call to get_map, which caused us to see 61 calls instead.

To fix this, we start the test off with AAE disabled, and then enable it later on once we're passed the race-prone part of the test (we do need to turn it on later, and can't just leave it disabled, since the end of the test requires AAE). This ended up being somewhat tricky, since the yokozuna_rt:rolling_upgrade function that this test uses was previously hardcoded to enable AAE on the upgraded cluster. This particular feature of the function seems to be safe to remove though, and Jason recently wrote a new, refactored version of the rolling_upgrade function as well, so I've incorporated that into the fixes.

On a final, unrelated note, I also removed an extra call to rt:clean_cluster that was previously at the end of the test. Just seemed unnecessary, since the whole environment gets cleaned up anyway when we start a new test, and I figured sometimes we may even want things left as-is after the test finishes if we're investigating an error.

### DIFF
--- a/src/yokozuna_rt.erl
+++ b/src/yokozuna_rt.erl
@@ -31,7 +31,6 @@
          override_schema/5,
          remove_index_dirs/2,
          rolling_upgrade/2,
-         rolling_upgrade/3,
          search/4,
          search/5,
          search_expect/5,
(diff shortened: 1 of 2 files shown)


zeeshanlakhani:
+1 on 293a4c95fef3514885a64945df651ef503f5876d. Waiting for @fadushin to test it once.

PR #1032, opened 2016-04-01
Add ts_simple_recompile_ddl

javajolt:
This test verifies that tables with mismatched DDL versions or old (i.e. missing) DDL compiler versions are actually recompiled.  The greatest difficulty in verifying this test was reading the results from the DETS table.  It turns out the trick is that the default DETS write-to-disk interval is 3 minutes, to by adding a manual `dets:sync` with each write, the test became much simpler.

Requires https://github.com/basho/riak_kv/pull/1377 and https://github.com/basho/riak_ql/pull/115

### DIFF
--- a/tests/ts_simple_recompile_ddl.erl
+++ b/tests/ts_simple_recompile_ddl.erl
@@ -55,6 +55,11 @@ simulate_old_dets_entries() ->
     Table1DDL = sql_to_ddl(create_table_sql("Table1")),
     Table2DDL = sql_to_ddl(create_table_sql("Table2")),
     Table3DDL = sql_to_ddl(create_table_sql("Table3")),
+
+    %% Here we want to test 3 degenerate cases:
+    %% 1) An old DETS entry (pre-1.3) which does not a DDL compiler version
+    %% 2) A compiled table with an older (pre-1.3) version
+    %% 3) A table which seemingly was stuck in the compiling state
     ok = dets:insert(?TABLE, {<<"Table1">>, Table1DDL, Pid, compiled}),
     ok = dets:insert(?TABLE, {<<"Table2">>, 1, Table2DDL, Pid2, compiled}),
     ok = dets:insert(?TABLE, {<<"Table3">>, 1, Table3DDL, Pid3, compiling}),
--- a/tests/ts_simple_recompile_ddl.erl
+++ b/tests/ts_simple_recompile_ddl.erl
@@ -27,6 +27,7 @@
 -export([confirm/0]).
 -define(TABLE, ?MODULE).
 -define(DETS_TABLE, riak_kv_compile_tab_v2).
+-define(LEGACY_TABLE, riak_kv_compile_tab).
 
 confirm() ->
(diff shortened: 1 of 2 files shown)


tests/ts_simple_recompile_ddl.erl
```diff
@@ -0,0 +1,87 @@
...
+-define(TABLE, ?MODULE).
+-define(KV_MODULE, riak_kv_compile_tab).
+
+confirm() ->
+    {Cluster, _Conn} = ts_util:cluster_and_connect(single),
+    Node = hd(Cluster),
+    lists:foreach(
+        fun(Table) ->
+            DDL = create_table_sql(Table),
+            ts_util:create_and_activate_bucket_type(Cluster, DDL, Table)
```
hmmr: In the new test modules, why not use the direct "CREATE TABLE" query rather than two separate, `riak-admin` based commands?

javajolt: Great question. Perhaps because it wasn't there originally when the testing infrastructure was built?  It would be an easy change, but let's do that in a separate PR.

hmmr: Agree. Possibly, also consider adding a subtest to ts_cluster_create_table_pass to check the equivalence of creating a table via query vs creating table via riak-admin.

tests/ts_simple_recompile_ddl.erl
```diff
@@ -0,0 +1,87 @@
...
+    open_dets(),
+    Pid = spawn_link(fun() -> ok end),
+    Pid2 = spawn_link(fun() -> ok end),
+    Pid3 = spawn_link(fun() -> ok end),
+    Table1DDL = sql_to_ddl(create_table_sql("Table1")),
+    Table2DDL = sql_to_ddl(create_table_sql("Table2")),
+    Table3DDL = sql_to_ddl(create_table_sql("Table3")),
+    ok = dets:insert(?TABLE, {<<"Table1">>, Table1DDL, Pid, compiled}),
+    ok = dets:insert(?TABLE, {<<"Table2">>, 1, Table2DDL, Pid2, compiled}),
+    ok = dets:insert(?TABLE, {<<"Table3">>, 1, Table3DDL, Pid3, compiling}),
```
hmmr: That `compiling` atom you are inserting last, isn't it going to make the `verify_resulting_dets_entries` fail?

Generally, some comments here would help reviewers to understand the underlying logic of the test. As it is, in `verify_resulting_dets_entries` you are reading the entries you have just inserted (in `simulate_old_dets_entries`), which is bound to succeed.

Aha, got it. The `'compiling'` will be turned to `'compiled'` in the process of node going up. I'm smart, no?

hmmr:
+1 1389ab6

tests/ts_simple_recompile_ddl.erl
javajolt: You are, indeed, smart.  Let me add some comments on the 3 cases being tested.

hmmr:
+1 3240f09

macintux:
:+1:

PR #1036, opened 2016-04-07
RTS-1040: Fix the return arguments from riak_shell:loop_TEST

javajolt:
Looks like this code path got overlooked


hmmr:
+1 8b20787

PR #1038, opened 2016-04-08
Compile ts_simple_create_table_eqc only if EQC is present

javajolt:
This test won't compile without EQC installed


javajolt:
Just yank the test for now, until it's done.

PR #1010, opened 2016-03-04
Flexible keys

andytill:
Tests for https://github.com/basho/riak_kv/pull/1357,  https://github.com/basho/riak_ql/pull/108

PR #1040, opened 2016-04-11
Tests For Flexible Keys 2

andytill:
This PR https://github.com/basho/riak_test/pull/1010 but without the ct integration and non-relevant commits.


tests/ts_cluster_keys_SUITE.erl
```diff
@@ -0,0 +1,722 @@
...
+    "CREATE TABLE table1 ("
+    "a SINT64 NOT NULL, "
+    "b SINT64 NOT NULL, "
+    "c TIMESTAMP NOT NULL, "
+    "d SINT64 NOT NULL, "
+    "PRIMARY KEY  ((a,b,quantum(c, 1, 's')), a,b,c,d))".
+
+select_exclusive_def_1_test(Ctx) ->
+    Query =
+        "SELECT * FROM table1 WHERE a = 1 AND b = 1 AND c > 0 AND c < 11",
```
lehoff: I would consider creating a function to return these query strings to make sure the arguments for the results and the query string always are in sync:
```
query_range_string(Start, End, StartInclusive, EndInclusive) ->
    io_lib:format("SELECT * FROM table1 WHERE a = 1 AND b = 1 AND c ~p ~B AND c ~p ~B",
                         lists:flatten([start_incl(StartInclusive, Start), end_incl(EndInclusive, End+1)]).

start_incl(true, Start) -> [">=", Start];
start_incl(false, Start) -> [">", Start-1].

end_incl(true, End) -> ["<=", End];
end_incl(false, End) -> ["<", End+1].
```

And the same for the results:
```
results_list(Start, End) ->
    [{1, 1, N, 1} || N <- lists:seq(Start, End)].
```

Which allows us to write a generic test function:
```
generic_select_test_args(Start, End, StartIncl, EndIncl, Ctx) ->
    Query = query_range_string(Start, End, StartIncl, EndIncl),
    Results = results_list(Start, End),
    {{column_names_def_1(), Results},
       riakc_ts:query(client_pid(Ctx), Query)}
```

And then a test can be written like this:
```
select_exclusive_def_1_test(Ctx) ->
    {Exp, Res} = generic_select_test_args(1,10, false, false, Ctx),
    ts_util:assert_row_sets(Exp, Res).
```

The final comparison has to be in the test to ease finding the problem in my experience, but there could be smarter ways.

Not necessary to it exactly as I have written - just the main idea of making the query string and the results align based on arguments instead of knowing it. The best way of doing that is the desired outcome.

lehoff:
I think it would be good to have a test where the primary key re-uses a field in the quantum:
```
PRIMARY KEY  ((a,c,quantum(c, 1, 's'))
```

tests/ts_cluster_keys_SUITE.erl
lehoff: Had a discussion with @andytill and we now agree that it becomes harder to read the tests when using the functions, so we leave it as is. 
@andytill will do an experiment on having `riakc_ts:query(client_pid(Ctx), Query)` in a helper function - if is more readable it goes in otherwise he just replies to this comment with a "Not better".

javajolt:
@andytill Are you planning to address @lehoff's comments?

andytill:
@javajolt yes I have one PR up https://github.com/basho/riak_ql/pull/119 and another coming in kv.

javajolt:
@andytill OK, I'll wait for the other PRs to be completed before looking at this in anger.

lehoff:
+1 
Ready to be merged.

PR #1042, opened 2016-04-11
Added EQC tests for CREATE TABLE

gordonguthrie:
This PR requires https://github.com/basho/riak_ql/pull/118


javajolt:
Because this added the dependency upon `riak_ql`, look for PRs soon to remove that dependency.

javajolt:
Actually, let's just see if we can remove that inclusion of the QL header.  That'll make this much more isolated and more maintainable.

javajolt:
@gordonguthrie Now I remember.  We didn't include this PR because it requires the inclusion of `riak_ql/include/riak_ql_ddl.hrl` which requires QL to actually be include in `riak_test` which is not what I think we want to do.  We can **use** QL, specifically the library in the devrel under test, but we don't want to tie it to the riak_test build.  

What I wanted to do is add some helper function in QL to build some of the data structures which currently require the QL header.

Make sense?

gordonguthrie:
Obsoleted

javajolt:
What replaced this?

PR #1044, opened 2016-04-15
ts_simple_recompile_ddl: Wait for cluster after restarting the node

javajolt:
See if pausing before reading the results helps to get a consistent test result


macintux:
:+1:

PR #1047, opened 2016-04-17
assorted post TTB merge fixes

hmmr:
This PR contains assorted fixes to correspond with the changes in https://github.com/basho/riak_kv/pull/1396 and https://github.com/basho/riak_kv/pull/1397. In particular:

* error messages are now more human-readable and don't include odd `{MnemonicAtom, ErrorDescription}` forcibly converted to a string using "~p";
* `riakc_ts:get_coverage` now returns native Elang term, not a `#tscoverageresp` that needs to be disassembled using internal PB records (they should not be exposed to the callers of `riakc_ts:get_coverage`).

### DIFF
--- a/tests/ts_cluster_comprehensive.erl
+++ b/tests/ts_cluster_comprehensive.erl
@@ -56,7 +56,7 @@ run_tests(PvalP1, PvalP2) ->
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3,
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3,
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3]),
-    ts_util:create_and_activate_bucket_type(Cluster, lists:flatten(TableDef), ?BUCKET),
+    ?assertEqual({[], []}, riakc_ts:query(rt:pbc(hd(Cluster)), TableDef)),
 
     %% Make sure data is written to each node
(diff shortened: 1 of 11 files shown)


tests/ts_cluster_comprehensive.erl
```diff
@@ -56,7 +56,7 @@ run_tests(PvalP1, PvalP2) ->
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3,
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3,
                   ?PKEY_P1, ?PKEY_P2, ?PKEY_P3]),
-    ts_util:create_and_activate_bucket_type(Cluster, lists:flatten(TableDef), ?BUCKET),
+    ?assertEqual({ok, {[], []}}, riakc_ts:query(rt:pbc(hd(Cluster)), TableDef)),
```
javajolt: Seems like you could call `ts_util:cluster_and_connect` above to get a cluster and a connection, then use the connection here instead of calling `rt:pbc` directly.  It also verifies that the connection is in place.

javajolt:
Getting this in to get the tests to pass

javajolt:
👍 9615800

tests/ts_cluster_comprehensive.erl
hmmr: I meant to create the table via a CREATE query, not via `riak-admin` two-step procedure. Either way, it's all the same.

PR #1049, opened 2016-04-18
Adjust STDDEV tests to match the current definition

javajolt:
According to the code and the documentation, STDDEV =:= STDDEV_SAMP, not STDDEV_POP, so make sure the test is actually doing the right thing

### DIFF
--- a/tests/ts_util.erl
+++ b/tests/ts_util.erl
@@ -504,7 +504,7 @@ get_optional(N, X) ->
     end.
 
 
--define(DELTA, 1.0e-10).
+-define(DELTA, 1.0e-15).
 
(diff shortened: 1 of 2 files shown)


tests/ts_util.erl
```diff
@@ -504,7 +504,7 @@ get_optional(N, X) ->
     end.
 
 
--define(DELTA, 1.0e-10).
```
andytill: Does this mean the results have to be more more accurate? And this would have uncovered the bug in the test?

javajolt: Yes, the numbers between the two standard deviation calculations are close but because they are not an exact match I wanted to tighten up the range.  Sadly comparing floats is not an exact science.

andytill:
+1 eb4bd37

PR #1054, opened 2016-04-21
Add confirm_missing to rt: and use it in ts_replication

javajolt:
Added to remove 10-minute delay in `ts_replication` test.


javajolt:
👍 7c62526

PR #1057, opened 2016-04-26
ts_simple_http_api: tolerate initial leveldb backend timeouts

hmmr:
At startup, riak_kv is ready before the eleveldb backend effectively is. This causes sporadic timeouts to occur in wm callbacks, resulting in 503 http responses. A simple workaround is proposed whereby a request is repeated 3 times before reporting a timeout failure.

### DIFF
--- a/tests/ts_simple_http_api_SUITE.erl
+++ b/tests/ts_simple_http_api_SUITE.erl
@@ -38,6 +38,11 @@ suite() ->
 init_per_suite(Config) ->
     [Node|_] = Cluster = ts_util:build_cluster(single),
     rt:wait_for_service(Node, riak_kv),
+    %% do a test query to overcome any initial eleveldb hiccups
+    URL = query_url(Node),
+    ?assertNotEqual(
+       {error, timeout},
+       patient_req(URL, [], post, "describe table fafa", 3)),
     [{cluster, Cluster} | Config].
 
 end_per_suite(_Config) ->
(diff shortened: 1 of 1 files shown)


andytill:
I'm not sure about this, retrying in all tests hid the problems with list keys coverage.  How about a section in the init suite that keeps attempting writes, or something that will return 503 and blocks until it succeeds, or times out. That means once if the init is successful then all tests can assume errors in responses are actually errors.

andytill:
Just seen that javajolt had done something similar for web machine here https://github.com/basho/riak_test/pull/1056/files

hmmr:
@andytill The idea of only allowing a request to fail with 503 once, initially, makes sense. It will have to be checked on a real leveldb operation (simply checking if webmachine app is up is not enough). Subsequent timeouts should be regarded as errors and crash the test.

I'll do the coding and push a commit shortly.

A little explanation: The particular 503 error is one of these: https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_wm_ts_util.erl#L233-L238, originating from timeouts happening here: https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_wm_timeseries_query.erl#L205-212 (these are the only two places in riak_kv_wm_timeseries callbacks generating the 503 response). In the r_t environment, it only happens during startup, and is apparently caused by leveldb taking some time to warm up.

andytill:
Just mentioned that other PR because of the approach rather than it fixing the problem.

I wonder why this doesn't happen through the pb interface.

javajolt:
LOL. Sounds like Andy and I had some similar questions and concerns.

javajolt:
Good news: Test passes on OSS.
Bad news: Still fails on EE.

Would love to learn *why* we are consistently getting EE failures rather than just coding the test around the failures.

hmmr:
Looks like @javajolt got it to get past the initial 503 somehow. Closing this for now (but will leave the branch around, just in case).

PR #1059, opened 2016-04-27
Test partition keys when the quantum is not the last field

andytill:
Assert that rows with keys that do not contain values in the partition key are not included in the range scan when the quantum is not the last element in the partition key.

### DIFF
--- a/tests/ts_cluster_keys_SUITE.erl
+++ b/tests/ts_cluster_keys_SUITE.erl
@@ -801,3 +801,32 @@ select_on_quantum_second_table_three_fields_test(Ctx) ->
         {rt_ignore_columns,Results},
         run_query(Ctx, Query)
     ).
+
+select_on_quantum_second_table_with_filters_test(Ctx) ->
+    %% setup
+    ?assertEqual(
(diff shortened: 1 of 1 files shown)


andytill:
Got a failing test for this.

```
----------------------------------------------------
2016-04-27 14:03:53.776
ROW DIFF
........
+ [{3,1000,<<"y">>,<<"w">>}]
.
+ [{3,1100,<<"y">>,<<"w">>}]
.
+ [{3,1200,<<"y">>,<<"w">>}]
.
+ [{3,1300,<<"y">>,<<"w">>}]
.
+ [{3,1400,<<"y">>,<<"w">>}]
.
+ [{3,1500,<<"y">>,<<"w">>}]
```

Closing this PR and putting validation in to ensure that the quantum is last, if at all.

javajolt:
@andytill So based on your mumble comments, we are just reverting back to having the quantum be the last field

andytill:
@javajolt yep, I will add the validation back into riak_ql, but the query stuff can stay until it is reviewed by the product team for use cases.

PR #1060, opened 2016-04-27
Remove flexible quantum tests

andytill:
Remove tests which assert on tables where the quantum is not the last element in the partition key, since this is no longer allowed.


javajolt:
👍

PR #1064, opened 2016-04-30
RTS-1087: Rename expected strings from "riak_shell" to "riak-shell"

javajolt:
When the strings were renamed in https://github.com/basho/riak_shell/pull/32, we needed to change the expected results to match.


hmmr:
+1 7e88038

PR #1069, opened 2016-05-04
RTS-1088: Start ibrowse when beginning the HTTP test

javajolt:
Initial experiments are promising that this helps to pass the `ts_simple_http_api_SUITE` test


javajolt:
This replaces #1056

lehoff:
+1

PR #1071, opened 2016-05-06
Add repl code to test deletes

nickelization:
Up until now, none of the riak_test modules for repl have tested deletes. These changes add some code to verify that deletions are propagated across clusters for both realtime and fullsync replication.

### DIFF
--- a/tests/repl_util.erl
+++ b/tests/repl_util.erl
@@ -13,6 +13,7 @@
          wait_until_connection/1,
          wait_until_no_connection/1,
          wait_for_reads/5,
+         wait_for_deletes/5,
          wait_until_fullsync_started/1,
          wait_until_fullsync_stopped/1,
          start_and_wait_until_fullsync_complete/1,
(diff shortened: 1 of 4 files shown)


tests/repl_util.erl
```diff
@@ -195,9 +196,15 @@ wait_until_fullsync_stopped(SourceLeader) ->
                   end).
 
 wait_for_reads(Node, Start, End, Bucket, R) ->
+    wait_for_x(Node, fun() -> rt:systest_read(Node, Start, End, Bucket, R, <<>>, true) end).
+
+wait_for_deletes(Node, Start, End, Bucket, R) ->
+    wait_for_x(Node, fun() -> rt:systest_verify_delete(Node, Start, End, Bucket, R) end).
+
+wait_for_x(Node, Fun) ->
```
JeetKunDoug: Maybe "wait_for_empty_result" would be a more appropriate name - rename "Reads" to "Results" as well.

JeetKunDoug:
:+1:

PR #1072, opened 2016-05-10
Streaming Queries Tests

andytill:
Tests for PRs:
- https://github.com/basho/riak_kv/pull/1415
- https://github.com/basho/riak-erlang-client/pull/281
- https://github.com/basho/riak_test/pull/1072

### DIFF
--- a/tests/ts_util.erl
+++ b/tests/ts_util.erl
@@ -139,9 +139,9 @@ ts_insert(Conn, Table, Columns, Data) ->
     ValClause = string:strip(lists:foldl(TermFn, [], tuple_to_list(Data)), right, $,),
     SQL = flat_format("INSERT INTO ~s (~s) VALUES (~s)",
                       [Table, ColClause, ValClause]),
-    lager:info("~ts", [SQL]),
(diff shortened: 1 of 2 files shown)


tests/ts_util.erl
```diff
@@ -138,19 +139,19 @@ ts_insert(Conn, Table, Columns, Data) ->
     ValClause = string:strip(lists:foldl(TermFn, [], tuple_to_list(Data)), right, $,),
     SQL = flat_format("INSERT INTO ~s (~s) VALUES (~s)",
                       [Table, ColClause, ValClause]),
-    lager:info("~ts", [SQL]),
+    % lager:info("~ts", [SQL]),
```
javajolt: How about just renaming this `lager:debug`?

andytill: Ooops. I didn't mean to commit that, do we need to log though? I think this is a side effect of building the SQL dynamically instead of plain text.  The plain text bulks up the test build will not be read as often as the test logs.

andytill:
Closing for now. Feature not expected in 1.6.

PR #1079, opened 2016-05-31
Fix bug in replication2 delete tests

nickelization:
This code worked fine if replication2 is run as a standalone test, but replication2 is also used as part of other bigger tests, and it can fail. The problem is that keys 1-100 are not necessarily always written, depending on how the test is being run. By using the 101-200 keys to test deletions instead, we can avoid this failure case.


JeetKunDoug:
+1 - replication2_upgrade passed w/ this fix in place, and code review makes sense.

PR #1081, opened 2016-06-02
new module to check for lk2pk conversion fun in list_keys code path

hmmr:
RTS-1126

A new module to test lk2pk conversion fun in list_keys code path, specifically as fixed in https://github.com/basho/riak_kv/pull/1427.

### DIFF
--- a/tests/ts_cluster_list_irreg_keys_SUITE.erl
+++ b/tests/ts_cluster_list_irreg_keys_SUITE.erl
@@ -30,7 +30,7 @@
 %% b, c, ts, d], but with different participation of the first four in
 %% the primary key. The atoms are mnemonics for the field order in the
 %% key.
--define(ALL_VARIANTS, [abct, acbt, bat]).
+-define(ALL_VARIANTS, [abct, acbt, abtc, bat]).
 
 -define(LIFESPAN, 250).  %% two and a half quanta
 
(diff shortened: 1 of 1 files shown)


erikleitch:
Test module looks good.  Appropriately fails in TS1.3 codebase, passes with corresponding branch of riak_kv (this is a companion to https://github.com/basho/riak_kv/pull/1427).

My only suggestion would be, for completeness' sake, to add a test where the local key includes additional fields that are not in the partition key.

hmmr:
@erikleitch It won't work I'm afraid. A DDL with a PK/LK defined like this: `primary key ((a, b, quantum(ts, 100, s)), a, b, c, ts)` is invalid (error 1020, "Local key does not match partition key").

erikleitch:
@hmmr 

Yes: 

primary key ((a, b, quantum(ts, 100, s)), a, b, c, ts) 

is invalid (because the local key must include _at least_ the partition key fields, and in the same order), but

primary key ((a, b, quantum(ts, 100, s)), a, b, ts, c)

(for example) is valid.  I'm just suggesting that we include (valid) examples of the additional case where there are extra fields in the local key that aren't in the partition key, against future changes that may break that functionality.

hmmr:
Added a test case as suggested by @erikleitch.

erikleitch:
+1  09a79bd

hmmr:
@borshop merge

PR #1082, opened 2016-06-03
Fix verify_kv_health_check test

jvoegele:
The health check feature has been deprecated but this test remains for backwards compatibility. Fix the test so that it enables health checks and that it actually checks the results of the `wait_until` calls used
in the test.


JeetKunDoug:
:+1:

javajolt:
@jvoegele I assume we'll be merging all of these test fixes forward, too?

PR #1083, opened 2016-06-06
Common Test suite for security and permissions RTS-978

andytill:
Testing create table and puts.


hmmr:
It would be appropriate to have a test for an attempt to do the same operation after permission has been revoked.

And, incidentally, does a permission record (say, "user user1 has permission to create table One") survive the deletion of the user (`riak-admin security del-user user1`)?

andytill:
@hmmr the security system is tested in pb_security.erl, this just tests that permissions are working correctly for TS actions.  I have written current progress on the JIRA. This suite needs a lot more tests to cover everything.

hmmr:
+1 0060998

PR #1084, opened 2016-06-07
Adds a "brutal kill node and check AAE" test.

JeetKunDoug:
- With appropriate anti-entropy limits (updated) the test should still pass even with a brutally killed node in the mix.
- Written to add confidence that AAE will, in fact, resolve issues where we previously saw this test fail. It used to fail because we didn't give it enough time/AAE concurrency in riak_kv to complete the KV side of the data repair. Now, with higher limits, it will converge to the correct values.


fadushin:
+1 f630384

PR #1085, opened 2016-06-09
Update riak_control to fix issues with not waiting for negotiaions

JeetKunDoug:
Negotiations of capabilities must complete before results are valid. Use rt:wait_until in several places to allow cluster to realize it's all ready to go. Remove invalid results from valid_status. Fix lager:info call.


fadushin:
+1 0e3d59c

PR #1088, opened 2016-06-14
ts_cluster_comprehensive: don't test overwriting TS data (because w1c)

hmmr:
Because TS data are write-once, and because overwriting the data will result in, eventually, a random pick from the siblings for a subsequent get, the existing test of the overwriting put in ts_cluster_comprehensive, makes little sense. I am now correcting my assumption of long ago that it should always work. It leaves one wondering why it never failed.


macintux:
If data is sent through a different coordinating node, or if a node is unavailable, more likely to see this test failure. In any event, we make no guarantees about behavior on overwrite of a w1c value, so agree this is the best thing to do.

macintux:
:+1: ff0aad2

PR #1089, opened 2016-06-14
Change verify_build_cluster reads to use r=3

nickelization:
We're hypothesizing that this will fix some intermittent errors we're
seeing in this test, where we get several notfound results after taking
down node 2. If two replicas end up on node 2 after we kill node 1, then
taking down node 2 would leave only one replica left, so an r=2 read
could return notfound. This matches up with our findings that subsequent
reads return the correct data (suggesting that the first read attempt is
triggering read repair).


bsparrow435:
+1 0d84d846b59f51075e71ded69685f7a106203f51

nickelization:
Doug also says it now passes on his test instance that was previously failing consistently - looks like we've got ourselves a fix.

javajolt:
@nickelization do we have plans to merge this forward?

nickelization:
@javajolt yeah we should definitely do that. What's the merge strategy for riak_test these days?

PR #1090, opened 2016-06-15
Test timeseries bugfix for realtime queue activity

macintux:
Test to validate basho/riak_repl#744 (RIAK-1588) (RIAK-1859) (RIAK-1864).

Fails under 1.3, passes with repl patch applied.

### DIFF
--- a/tests/ts_replication.erl
+++ b/tests/ts_replication.erl
@@ -3,6 +3,8 @@
 -export([confirm/0]).
 -include_lib("eunit/include/eunit.hrl").
 
+-compile({parse_transform, rt_intercept_pt}).
+
 %% Test both realtime and fullsync replication for timeseries data.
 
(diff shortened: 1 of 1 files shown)


tests/ts_replication.erl
```diff
@@ -104,6 +106,22 @@ replication(ANodes, BNodes, Table, NormalType) ->
     create_normal_type(ANodes, NormalType),
     create_normal_type(BNodes, NormalType),
 
+    %% Make certain we don't start throwing objects into the realtime
```
JeetKunDoug: Should this same intercept/test be done in other places (fullsync test, for example, to make sure we're not pushing data if fullsync is `both` or `fullsync`?

JeetKunDoug:
+1 for now - would like to revisit testing the other cases though.

PR #1091, opened 2016-06-15
A rolling upgrade/downgrade test for TS

hmmr:
RTS-1155

A rolling upgrade-downgrade test for a 3-node cluster, verifying data retrieval (via a SELECT query) from each node in a fully upgraded, fully downgraded, and mixed cluster. The scenarios covered are described in https://github.com/basho/rfc/blob/master/support/upgrade_downgrade_specs.md.

The cluster is created from nodes at the 'current' version; a table is created and sample data is inserted. The test proceeds to run a battery of SELECT queries on it, in six iterations. One node is downgraded to the 'previous' version (as designated in ~/.riak_test.config) after each iteration, upon which the nodes are upgraded back to 'current'.

In each iteration, a total of 10 SELECT queries are executed on each node, featuring assorted aggregation functions.

The client used to issue query calls to the cluster will be of the appropriate version (either 'current' or 'previous'). A modified version of standard OTP/common_test/ct_slave facility will be used for the 'previous' client.

**update:** The test is now expanded to
* create not one but two tables on each node, one when the node is upgraded, and another, when the node is downgraded;
* each table will be queried from all nodes, and both when the querying node is upgraded and when it it is downgraded;
* each table will be queried when the node it was created on, is upgraded as well as downgraded.

Also, only the 'current' client is used for all queries (the selective use of current/previous client still can be undisabled in code).

**update:** The facility is reworked to be more eqc-friendly. `ts_updown_util:run_scenario/2` can now be used from eqc-enabled tests (a simple scenario generator is in ts_updowngrade_select_aggregation_SUITE.erl). For non-eqc setups, module writers can use `ts_updown_util:run_scenarios/2`

A scenario describing a test case is given in r_t module as a record with fields describing the requirements for the upgrade/downgrade test as follows:

```Erlang
-type version() :: current | previous.
-record(scenario, {
          %% riak version on the node where CREATE TABLE query will be issued
          table_node_vsn :: version(),

          %% riak version on the node where SELECT queries will be issued
          query_node_vsn :: version(),

          %% whether to up- or downgrade the interesting nodes after
          %% table creation (before running SELECT queries)
          need_table_node_transition :: boolean(),
          need_query_node_transition :: boolean(),

          %% hints to try to ensure cluster homogeneity after any node
          %% transitions (may not be honoured depending on whether the
          %% interesting nodes emerge in different or same version
          %% after transitions)
          need_pre_cluster_mixed :: boolean(),
          need_post_cluster_mixed :: boolean(),

          %% table name (useful to avoid name clashes since tables are
          %% not cleaned up after each test)
          table :: binary(),

          %% Table DDL and the data to write to it
          ddl :: binary(),
          data :: [row()],

          %% a list of {SelectQueryFmt, Expected} (`SelectQueryFmt'
          %% must contain a single "~s" placeholder for the table name
          select_vs_expected :: [{binary(), term()}]
         }).
```

Error reports will be collected in the form of this record
```Erlang
-record(failure_report, {
          %% node composition, with versions at each node
          cluster :: [{node(), version()}],

          %% node where table was created
          table_node :: node(),
          %% node where the failing SELECT was issued
          query_node :: node(),

          %% whether the table and querying nodes were upgraded or
          %% downgraded prior to SELECT query
          did_transition_table_node :: boolean(),
          did_transition_query_node :: boolean(),

          %% the failing query proper
          failing_query :: binary(),

          %% Expected and Got
          expected :: term(),
          error :: term()
         }).
```

c/o @javajolt, @gordonguthrie.


tests/ts_updown_util.erl
```diff
@@ -0,0 +1,361 @@
...
+         run_scenario/2
+        ]).
+
+-include_lib("eunit/include/eunit.hrl").
+-include("ts_updown_util.hrl").
+
+-type versioned_cluster() :: [{node(), version()}].
+%% preparations
+
+make_config(Config) ->
```
gordonguthrie: This seems to be misnamed - the config is a side-effect of building the cluster - maybe rename this function `build_cluster_and_and_return_config`

src/rt_slave.erl
```diff
@@ -0,0 +1,371 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc rt_slave is a modified version of ct_slave.erl from the
```
gordonguthrie: Are we totally sure that a fork is the way to go here? Can/should the fixes be posted to Ericsson/OTP team?

tests/ts_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,243 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(ts_updowngrade_select_aggregation_SUITE).
```
javajolt: I would suggest `ts_cluster_updowngrade_select_aggregation_SUITE` to match our test-naming pattern

javajolt: Or to simplify `ts_cluster_updowngrade_aggregation_SUITE`

src/rt_slave.erl
hmmr: Sending a patch upstream is definitely the way to finish off this isolated issue, but it's clearly a separate affair. For the case in point, forking the OTP version seems to be the least disruptive option.

hmmr:
Along with some renames done in c789faf to address @gordonguthrie and @javajolt comments, the branch has been rebased onto current riak_ts-develop.

src/rt_client.erl
```diff
@@ -0,0 +1,100 @@
...
+               [rtdev:relpath(previous),
+                "dev/dev1/lib/riakc-*/ebin"]))),
+    true = rpc:call(
+             SlaveNode, code, replace_path, ["riakc", PrevRiakcPath]),
+    SlaveNode.
+
+load(Version) ->
+    LoadPaths = get_load_paths(Version),
+    Modules = get_modules(LoadPaths, []),
+    %% ct:pal("switching client from vsn ~s (~s)", [current_client_vsn(), Version]),
```
gordonguthrie: These are useful ct:pal calls aren't they?

src/rt_client.erl
```diff
@@ -0,0 +1,100 @@
...
+     filename:join([Root, "dev/dev1/lib/riakc*/ebin"]),
+     filename:join([Root, "dev/dev1/lib/riak_pb*/ebin"])
+    ].
+
+%% we have a problem: the set of files in Version X of the client
+%% might not be congruent with that of Version Y
+%% so we will hard unload files from the app manifest
+%% and then iterate over the beam files in the path and unload any
+%% modules with those names - all a bit brute-force-and-ignorance
+unload_riak_client(Modules) ->
```
gordonguthrie: There remain problems with this in the sense that if riak-erlang-client acquires a new dependency on a version upgrade that dependency needs to be added to the code path and started as well. This change is hard-coded to riak_pb and riakc

src/rt_slave.erl
gordonguthrie: So we are proposing to do that and resolve the fork here appropriately?

It is worth adding a comment to the README or this module?

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,243 @@
...
+
+-export([
+         suite/0,
+         init_per_suite/1,
+         end_per_suite/1,
+         all/0,
+         run_this_test/1
+        ]).
+
+-ifdef(EQC).
```
gordonguthrie: Not convinced this should be a common test - a suite of 1 tests - but if we are strategically going to move to CT it would make sense to have it so...

But I am not convinced that we are going in that direction - its not entirely clear what the point of this comment is, except to get other peoples opinions...

src/rt_client.erl
andytill: I thought we were going to have completely separate nodes for different client versions, and not load/unload code paths?

hmmr: Indeed, this is legacy code from early days. Separate clients are run from separate nodes. I should/will clean up in this module.

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,243 @@
...
+                   need_query_node_transition = NeedQueryNodeTransition,
+                   need_pre_cluster_mixed = NeedPreClusterMixed,
+                   need_post_cluster_mixed = NeedPostClusterMixed,
+                   table = Table,
+                   ddl = ts_util:get_ddl(aggregation, binary_to_list(Table))}).
+
+gen_version() ->
+    ?LET(A, oneof([current, previous]), A).
+
+gen_table() ->
```
gordonguthrie: You will get random crashes if you use gen_name to generate a table name because gen_name doesn't return a unique set - the answer is to not specify the table name in the scenario record at all (nor substitute it into the CREATE TABLE, INSERT INTO or SELECT queries until run-time - and then use a timestamp-to-string...

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,243 @@
...
+    ?LET(A, ts_sql_eqc_util:gen_name("aggregation_test_"), A).
```
gordonguthrie: You get a 'free' underscore with gen_name already - so this doubles them up...

tests/ts_sql_eqc_util.erl
```diff
@@ -0,0 +1,165 @@
...
+%%      testing purposes.
+
+-module(ts_sql_eqc_util).
+
+-compile(export_all).
+
+-ifdef(EQC).
+-include_lib("eqc/include/eqc.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("../riak/deps/riak_ql/include/riak_ql_ddl.hrl").
```
javajolt: We do not want to create a hard-dependency upon `riak_ql` at compile time, due to how we don't lock `riak_test` modules.  We should add some testing functions to QL to allow it to be linked in at run time.  Happy to add these as necessary.

tests/ts_updown_util.erl
```diff
@@ -0,0 +1,372 @@
...
+    %% 0. retreive scenario-invariant data from Config
+    [Data, DDLFmt, SelectVsExpected] =
+        [begin
+             case Supplied of
+                 undefined ->
+                     ?CFG(Item, Config);
+                 Defined ->
+                     Defined
+             end
+         end || {Supplied, Item} <- [{Data_, data}, {DDL_, ddl},
```
gordonguthrie: Tagged tuples where the tag is the second element is just weird...

tests/ts_updown_util.erl
```diff
@@ -0,0 +1,372 @@
...
+        find_or_make_node_at_vsn(NodesAtVersions0, TableNodeVsn, []),
+    {QueryNode, NodesAtVersions2} =
+        find_or_make_node_at_vsn(NodesAtVersions1, QueryNodeVsn, [TableNode]),
+
+    %% 3. Try to ensure cluster is (not) mixed as hinted but keep the
+    %%    interesting nodes at their versions as set in step 1.
+    NodesAtVersions3 =
+        ensure_cluster(NodesAtVersions2, NeedPreClusterMixed, [TableNode, QueryNode]),
+
+    %% 4. Create table, put data (this step is always assumed to
```
gordonguthrie: This precondition isn't true though - the ASC/DESC tests for 1.4 MUST fail on CREATE TABLE...

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,205 @@
...
+        _ ->
+            ct:fail("There were failing queries", [])
+    end.
+
+prop_scenario(Config) ->
+    ?FORALL(
+       Sce, gen_scenario(),
+       [] == ts_updown_util:run_scenario(Config, Sce)).
+
+gen_scenario() ->
```
gordonguthrie: Not sure this generator has the appropriate shrinking properties - the actual cluster configuration that the test scenario is run against has hysteresis, it depends on the total state that results after running all the previous scenarios - if the scenario list is bisected (for instance) a particular scenario can be run in a different context. The version that I wrote generated cluster configurations - so we were modelling the ring - and the transitions were the delta - and thus deterministic - you are generating the transitions and the cluster state is only deterministic on application of the set of transitions...

gordonguthrie: What happens when this test fails? How does it shrink?

gordonguthrie: The generator I wrote in Ukraine had left and stopped states - so it would test a range of other ring states - these have been dropped in this implementation...

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
```diff
@@ -0,0 +1,205 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(ts_cluster_updowngrade_select_aggregation_SUITE).
+
```
gordonguthrie: The test itself is only the functions prop_scenario/1 and make_scenario_invariants/1 so the rest should be an upgrade/downgrade behaviour...

(Given the other problems with the generator kinda ignore this for now, its a comment...)

src/rt_slave.erl
hmmr: Filed a PR against basho/otp: https://github.com/basho/otp/pull/22.

src/rt_client.erl
hmmr: That code is gone.

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
hmmr: I believe all new tests should be CT-style. In this case, they have the advantage of preserving the list of failing queries in CT logs.

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
hmmr: Tables are now all dynamically generated, as you suggest.

tests/ts_sql_eqc_util.erl
hmmr: That module is removed (to be placed in riak_ql/tests).

tests/ts_updown_util.erl
hmmr: Well, they are not proplists but short-lived aggregates to be immediately decomposed within this list comprehension, but I can't help agreeing they look.. unnatural. Fixed in ba73dca.

tests/ts_updown_util.erl
hmmr: Fair point. Addressed in 02cc6de.

tests/ts_cluster_updowngrade_select_aggregation_SUITE.erl
hmmr: With respect to shrinking, it's a tough call. Because the parameters of the scenario (with or without taking into account the initial cluster configuration) are not scalar values, eqc logic cannot conceivably discard a region in the search space on the grounds of some parameter value causing a scenario to fail. For example, the observation that the table node transition from current to previous causes the subsequent query to fail, does not imply the same scenario where transition occurs in the opposite direction, should fail as well. Here, the parameter `need_table_node_transition` is `true` in both cases, and even so both scenarios need to be tested, independently.

The ideas of generating transitions randomly, chaining them to form multi-transition scenarios run on a cluster continuously, can be implemented in future (my preference still is to have dedicated, hard-coded scenarios for transitions other than stop-upgrade-start, because of combinatorial explosion of transitions -- 20 sec each). For the time being, the proposed static ladder should be "good enough" to emulate a typical cluster rolling upgrade.

hmmr:
Addressed all concerns, within a narrower scope of a static list of 64 scenarios, as we have agreed for now.

tests/ts_util.erl
```diff
@@ -522,30 +525,45 @@ get_optional(N, X) ->
...
+    assert(String, Exp, Got).
 
 assertf2([], []) -> pass;
-assertf2([H1 | T1], [H2 | T2]) ->
-    Diff = H1 - H2,
-    Av = (H1 + H2)/2,
-    if Diff/Av > ?DELTA -> fail;
-        el/=se          -> assertf2(T1, T2)
-    end.
+assertf2([H1 | T1], [H2 | T2]) when is_float(H1), is_float(H2) ->
```
javajolt: What was wrong with the numerical difference?  The sign?

hmmr: The sign was obviously not taken into account. ~~When I fixed that (`Diff = abs(H1-H2)`), I still had false negatives with values like `1788996260097066800.0` vs `1.788996260097067e18`, and decided to compare the strings, which worked.~~

After double-checking, it appears the abs fix is enough (the false negative was due to me taking some shortcuts in development process). I am reverting this commit now.

tests/ts_updown_util.erl
```diff
@@ -235,11 +197,11 @@ query_with_client(Query, Node, Config) ->
     case Version of
         current ->
             riakc_ts:query(Client, Query);
-        previous ->
-            rpc:call(
-              ?CFG(previous_client_node, Config),
-              riakc_ts, query, [Client, Query])
-    end.
+                              previous ->
```
hmmr: Unindent me please. (If you update emacs Erlang mode to the current version, it will stop treating `query` as a keyword, and will autoindent normally.)

gordonguthrie:
+1 e0075ba

hmmr:
@borshop merge

PR #1092, opened 2016-06-15
Added a test for an extractor which returns non-unicode data

fadushin:
And which will therefore result in a bad_data error when transcoding to JSON.


jvoegele:
+1 d82b34db8d2b044cc9c4ef29ca52b41fb00d8103

PR #1093, opened 2016-06-15
Update pipe_verify_handoff_blocking to handle vnode shutdown

JeetKunDoug:
Occasionally the test would fail with an error where a vnode was transferred after the vnode proxy had gotten its pid but before the work could be sent.


nickelization:
@JeetKunDoug There's a comment around line 110 that seems to refer to the old behavior involving `vnode_down`. Seems like it should be deleted.

nickelization:
Gotta run out for a bit, but other than that one comment, it all looks good to me. Once that's fixed, you can consider this +1'd.

JeetKunDoug:
After further code review, realized that the race condition mentioned isn't really what was going on - fixed the bug in riak_pipe:queue_work_wait in basho/riak_pipe#106. This test should fail without that fix, and pass with it.

PR #1095, opened 2016-06-20
Lock riakc version, since a strange version at HEAD will break giddyu…

ooshlablu:
…p runs

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -15,7 +15,7 @@
         {getopt, ".*", {git, "git://github.com/jcomellas/getopt", {tag, "v0.4"}}},
         {meck, ".*", {git, "git://github.com/basho/meck.git", {tag, "0.8.2"}}},
         {mapred_verify, ".*", {git, "git://github.com/basho/mapred_verify", {branch, "master"}}},
-        {riakc, ".*", {git, "git://github.com/basho/riak-erlang-client", {branch, "master"}}},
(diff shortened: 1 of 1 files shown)


hmmr:
+1 2385155

hmmr:
@javajolt I think this is in line with the latest pruning and grafting of branches in riak-erlang-client?

javajolt:
For 2.1 this is probably fine.  For 2.2+, however, we probably need a new tag.

rebar.config
```diff
@@ -15,7 +15,7 @@
         {getopt, ".*", {git, "git://github.com/jcomellas/getopt", {tag, "v0.4"}}},
         {meck, ".*", {git, "git://github.com/basho/meck.git", {tag, "0.8.2"}}},
         {mapred_verify, ".*", {git, "git://github.com/basho/mapred_verify", {branch, "master"}}},
-        {riakc, ".*", {git, "git://github.com/basho/riak-erlang-client", {branch, "master"}}},
+        {riakc, ".*", {git, "git://github.com/basho/riak-erlang-client", {tags, "2.1.2"}}},
```
nickelization: I think this should be `tag` not `tags`.

nickelization:
Once this issue is fixed, we'll need to merge a similar fix to the develop-2.2 and master branches.

ooshlablu:
@javajolt fixed in another pr

PR #1096, opened 2016-06-30
use rt:leave (not rt:stop) in ts_degraded_select_pass_* tests

hmmr:
Using `rt:leave` will force the handoff and actually check to ensure there is no data loss. The `ts_degraded_select_pass_*` used to just use `rt:stop`, which did not affect the cluster structure and was totally innocuous.


macintux:
:+1:

PR #1098, opened 2016-07-05
Fix the expected sub string in basic_command_line getpid test

andytill:
The error message is different for the `getpid` command, and the test was using `rt:str_mult/2` but not passing a list of strings.

Fixed by calling `rt:str/2` without the correct error message sub string.


hmmr:
+1 cc50b7c

Indeed, `rt:str_mult/2` takes a list of strings, and it should clearly be `rt:str` here.

PR #1100, opened 2016-07-05
Misconfigured test

gordonguthrie:
A set of 10 elements spanning multiple quanta is written to database
Then a query *inclusive* of the two end records was generated, but
the expected database was *exclusive* of the limits - causing it to fail


javajolt:
:+1: 4f2f63e

PR #1103, opened 2016-07-08
dependency cleanup

fadushin:
This PR pins dependencies on meck, riak erlanf client, and riak erlang http client to tagged dependencies, to reduce the chances that tests on this branch will fail because of floating dependencies.


javajolt:
+1 c8cf713

PR #1104, opened 2016-07-11
hll tests (against develop-2.2)

zeeshanlakhani:
Original (to master): https://github.com/basho/riak_test/pull/1063

- new test for HLL(set) datatypes
- update crdt cap and converge tests for hlls
- temp rebar updates for riak-client; *will update w/ correct version*

### DIFF
--- a/tests/test_hll.erl
+++ b/tests/test_hll.erl
@@ -1,0 +1,217 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2016 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%%% @doc r_t to test hll datatypes across a riak cluster
+
+-module(test_hll).
+
+-export([confirm/0]).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(HLL_TYPE1, <<"hlls1">>).
+-define(HLL_TYPE2, <<"hlls2">>).
+-define(HLL_TYPE3, <<"hlls3">>).
+-define(BUCKET1, {?HLL_TYPE1, <<"testbucket1">>}).
+-define(BUCKET2, {?HLL_TYPE2, <<"testbucket2">>}).
+-define(DEFAULT_P, 14).
+-define(SET_P, 16).
+-define(BAD_P, 1).
+-define(P_SETTING, hll_precision).
+-define(KEY, <<"flipit&reverseit">>).
+-define(CONFIG,
+        [
+         {riak_core,
+          [{ring_creation_size, 8},
+           {anti_entropy_build_limit, {100, 1000}},
+           {anti_entropy_concurrency, 8},
+           {handoff_concurrency, 16}]
+         }
+        ]).
+
+confirm() ->
+    %% Configure cluster.
+    Nodes = [N1, N2, N3, N4] = rt:build_cluster(4, ?CONFIG),
+
+    NodeRand1A = rt:select_random([N1, N2]),
+    NodeRand1B = rt:select_random([N1, N2]),
+    NodeRand2A = rt:select_random([N3, N4]),
+    NodeRand2B = rt:select_random([N3, N4]),
+
+    lager:info("Create PB/HTTP Clients from first two nodes, and then"
+               " the second two nodes, as we'll partition Nodes 1 & 2 from"
+               " Nodes 3 & 4 later"),
+
+    %% Create PB connection.
+    PBC1 = rt:pbc(NodeRand1A),
+    PBC2 = rt:pbc(NodeRand2A),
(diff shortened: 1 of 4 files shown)


tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+    PBC = rt:pbc(NodeRand1),
+    riakc_pb_socket:set_options(PBC, [queue_if_disconnected]),
+
+    %% Create HTTPC connection.
+    HttpC = rt:httpc(NodeRand2),
+
+    rt:create_and_activate_bucket_type(NodeRand1,
+                                       ?HLL_TYPE1,
+                                       [{datatype, hll},
+                                        {hll_precision, ?SET_P}]),
```
russelldb: Why not use the macro ?P_SETTING that you defined above here?

tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%%% @doc r_t to verify changes to serialization/metadata/changes
```
russelldb: I don't think this docstring is accurate

tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+    check_precision_and_reduce_test(C, CMod, Bucket, ?SET_P, HllSet0),
+
+    partition_write_heal(C, CMod, Bucket, Nodes),
+
+    HllSet1 = get_hll(C, CMod, Bucket),
+    check_precision_and_reduce_invalid_test(C, CMod, Bucket, ?SET_P - 1, HllSet1),
+
+    ok.
+
+add_tests(C, CMod, Bucket, NewS) ->
```
russelldb: Is it called `NewS` 'cos it is an `S` for Set? Picky picky I know, but I like the name Set or HLL here that is what I transposed it to in my mind when reading the code. No biggy. Just sayin'

russelldb: OH I SEE, the code explicitly expects this to be a New Set since the first thing it does is add an element and check the cardinality is 1. I get it now.

russelldb: Then why not just create the HLL in this `add_tests` function, and reduce the arity/complexity/repetition?

tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+check_precision_and_reduce_test(C, CMod, Bucket, ExpP, HllSet) ->
+    {ok, Props0} = CMod:get_bucket(C, Bucket),
+    ?assertEqual(proplists:get_value(?P_SETTING, Props0), ExpP),
+    ?assertEqual(riak_dt_hll:precision(HllSet), ExpP),
+    ok = CMod:set_bucket(C, Bucket, [{?P_SETTING, ExpP - 1}]),
+    {ok, Props1} = CMod:get_bucket(C, Bucket),
+    ?assertEqual(proplists:get_value(?P_SETTING, Props1), ExpP - 1).
+
+check_precision_and_reduce_invalid_test(C, CMod, Bucket, ExpP, HllSet) ->
+    ?assertEqual(riak_dt_hll:precision(HllSet), ExpP),
```
russelldb: I'm sure we can have this discussion elsewhere, but why did you decide to call it `riak_dt_hll` but the module lives in riak_kv. Do we plan to move it to riak_dt in future?

tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+
+    lager:info("Write to one side of the partition"),
+    {ok, S0} = CMod:fetch_type(C, Bucket, ?KEY),
+    add_element(C, CMod, Bucket, S0, <<"OH hello there">>),
+    {ok, S1} = CMod:fetch_type(C, Bucket, ?KEY),
+    ?assertEqual(riakc_hll:value(S1), 5),
+
+    lager:info("Heal"),
+    ok = rt:heal(PartInfo),
+    ok = rt:wait_for_cluster_service(Nodes, riak_kv).
```
russelldb: Shouldn't you then test that the new value is present after healing? Or shouldn't you have "incremented" on both sides and checked that the total was correct. I'm not clear on what is being tested here.

tests/test_hll.erl
```diff
@@ -0,0 +1,179 @@
...
+
+check_precision_and_reduce_test(C, CMod, Bucket, ExpP, HllSet) ->
+    {ok, Props0} = CMod:get_bucket(C, Bucket),
+    ?assertEqual(proplists:get_value(?P_SETTING, Props0), ExpP),
+    ?assertEqual(riak_dt_hll:precision(HllSet), ExpP),
+    ok = CMod:set_bucket(C, Bucket, [{?P_SETTING, ExpP - 1}]),
+    {ok, Props1} = CMod:get_bucket(C, Bucket),
+    ?assertEqual(proplists:get_value(?P_SETTING, Props1), ExpP - 1).
+
+check_precision_and_reduce_invalid_test(C, CMod, Bucket, ExpP, HllSet) ->
```
russelldb: Is this because HLL's can have their precision reduced but not increased? If so please add a comment that makes that explicit for the less educated reader.

russelldb:
A few questions/comments above, but otherwise, tests pass nicely, and it makes sense. Good work!

+1 (when the rest are +1'd obv!)

tests/test_hll.erl
zeeshanlakhani: Yeah, derpy on my part. Fixing/Fixed.

tests/test_hll.erl
zeeshanlakhani: Lager'ed for when the test runs.

russelldb:
Works for me

+1

PR #1105, opened 2016-07-14
RTS-1005: Fix tests for additional DESCRIBE columns

javajolt:
Adjust the tests to accommodate https://github.com/basho/riak_kv/pull/1438.  The test `ts_simple_http_api_SUITE` also does a `DESCRIBE` but compares the HTTP output to the TTB output so does not require a change.  The other tests just need the extra columns.

Also floating the Erlang client dependencies to `develop` since KV version has now merged with TS version.


paegun:
+1 37a2b6a

PR #1106, opened 2016-07-14
Merge develop-2.2 with riak-ts_develop

javajolt:
Get all the latest and greatest fixes


nickelization:
I suspect this might break some of the yz tests since the batching+draining work is not present in TS, but I talked to Brett and he said yz is not working right now anyway, so apparently that's not an issue. Reading through the rest of the changes, everything looks like it should be reasonably safe to merge. Not sure if any other tests might be broken by this merge, but according to Brett we're not expecting everything to pass at this stage, so I'll go ahead and give this a +1.

nickelization:
:+1: 09e9dea

PR #1107, opened 2016-07-15
RTS-1149: Add security tests for querying data

javajolt:
We already have a substantial test suite for testing PB security with TS.  Only a few additional cases were needed.

Also renamed the existing test suite to be `ts_simple_pb_security_SUITE` and added all the same tests to HTTP as `ts_simple_http_security_SUITE`.  It looks like HTTP still has issues with security since we've not yet added support which means this suite currently mostly fails.

### DIFF
--- a/tests/ts_simple_pb_security_SUITE.erl
+++ b/tests/ts_simple_pb_security_SUITE.erl
@@ -171,6 +171,19 @@ receive_keys(Pid, Acc) ->
         {ok, Acc}
     end.
 
+%% If a test creates a table, add it to the ever-growing laundry list
+%% of tables so those tests which need to know about all tables will
+%% have the complete list
+add_table_to_list(Config, Table) ->
+    {_Saver, ConfigList} = ?config(saved_config, Config),
+    {save_config, [{Table}|ConfigList]}.
+
+%% Just like add_table_to_list/2, but required to pass the the config
+%% state to the next test, even if there is no change
+add_no_table_to_list(Config) ->
+    {_Saver, ConfigList} = ?config(saved_config, Config),
+    {save_config, ConfigList}.
+
 %%--------------------------------------------------------------------
 %% TESTS
 %%--------------------------------------------------------------------
@@ -182,7 +195,8 @@ trusted_user_does_not_need_a_password_to_connect_test(Ctx) ->
         ["security", "add-user", User]),
     {ok,_} = riak_admin(Ctx,
         ["security", "add-source", "all", "127.0.0.1/32", "trust"]),
-    {ok, _} = client_pid(Ctx, User, NonsensePassword).
+    {ok, _} = client_pid(Ctx, User, NonsensePassword),
+    {save_config,[]}.
 
 password_user_cannot_connect_with_wrong_password_test(Ctx) ->
     User = "stranger",
@@ -194,7 +208,8 @@ password_user_cannot_connect_with_wrong_password_test(Ctx) ->
     ?assertEqual(
         {error,{tcp,<<"Authentication failed">>}},
         client_pid(Ctx, User, Password)
-    ).
+    ),
+    add_no_table_to_list(Ctx).
 
 with_security_user_cannot_create_table_without_permissions_test(Ctx) ->
     User = user_name(),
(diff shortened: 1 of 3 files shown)


andytill:
It is OK to have duplication in tests, tests are not production code. It is good to be able to clearly see what steps are being taken.

javajolt:
@andytill I agree that tests don't necessarily need to be optimized like production code, but I do find that refactoring boilerplate can:
- Ensure correct duplication (fewer copy pasta errors)
- More clearly define what each part of the code is doing, e.g., creating a table
- Remove the noise from individual tests to emphasize the differences

I learned a few things in the refactor, too.  It looked like there was duplication in some cases when they were subtly different.  By taking away the duplicated portions I could more clearly identify the differences.

However, if you feel strongly about this I can revert the refactoring and just add in my new tests.

andytill:
Yes, please revert the util function changes.

I understand your points, but in practice this approach has produced some tests that are very difficult to understand e.g. data is in util functions so it takes many look ups to work out what a test is doing, and tests being restricted by what util functions are available, trying to reuse util functions to keep tests small so they end up testing the same thing, and util functions accruing many arguments, or arities passing in defaults to the main function.

tests/ts_simple_pb_security_SUITE.erl
```diff
@@ -269,4 +268,62 @@ with_security_when_user_is_given_permissions_user_can_put_data_test(Ctx) ->
...
+    ),
+    ?assertEqual(
+        ok,
+        riakc_ts:put(Pid, <<"table6">>, [{1,1,1}])
+    ),
+    ?assertEqual(
+        {ok, {[<<"a">>, <<"b">>, <<"c">>], [{1,1,1}]}},
+        riakc_ts:query(Pid, "SELECT a, b, c FROM table6 WHERE a=1 AND b=1 AND c>0 and c<2")
+    ),
+    ?assertMatch(
```
andytill: Two different things being asserted in this test, queries and streaming list keys.

andytill:
Is this intended to be a comprehensive security test suite? If so, there are some missing tests. Each operation should have a test with and without security permissions, for PB and HTTP.

|       Feature       | PB with perms | PB no perms | HTTP with perms | HTTP no perms |
| ------------------- | ------------- | ----------- | --------------- | ------------- |
| Queries             | done          | done        | done            | done          |
| `CREATE TABLE`      | done          | done        | done            | done          |
| Streaming list keys | done          |             |                 |               |
| Client puts         | done          | done        | done            | done          |
| SQL `INSERT`        |               |             |                 |               |
| Coverage            |               |             |                 |               |
| Single Key Delete   |               |             |                 |               |
| Single Key Get      |               |             |                 |               |
| SQL `DESCRIBE`      |               |             |                 |               |

javajolt:
No, we already have tests for many other operations without security enabled for PB/TTB:
- `ts_cluster_coverage` - Coverage
- `ts_simple_get` - Single Key Get
- `ts_simple_single_key_ops` - Single Key Get, Single Key Put, Single Key Delete
- `ts_cluster_create_table_via_sql_SUITE` - SQL `DESCRIBE`
- `ts_simple_insert` - SQL `INSERT`

And for HTTP we have`ts_simple_http_api_SUITE` which covers:
- Delete (it's primary key, but does that mean single key?)
- SQL `DESCRIBE`
- Streaming List keys

We definitely should test each permission, however.  Also it looks like we have some holes in the HTTP coverage without security.  Let me see about plugging some of those gaps.

andytill:
> No, we already have tests for many other operations without security enabled for PB/TTB:

I mean where the client does have security permissions, so the request should be denied. I can't see any HTTP security tests apart from the ones you have added.

javajolt:
Now, I'm with you.  I found that list in `riak_kv_ts_api:api_call_to_perm`:
- `riak_ts.get`
- `riak_ts.put`
- `riak_ts.delete`
- `riak_ts.listkeys`
- `riak_ts.coverage`
- `riak_ts.query_create_table`
- `riak_ts.query_select`
- `riak_ts.query_describe`

javajolt:
Merging in https://github.com/basho/riak_kv/pull/1441 we are not using `riak_ts.query_insert` instead going with `riak_ts.put`.

@andytill I've added all of the **PB with perms** and **PB no perms** to the existing test suite.  I've not built out the HTTP tests since they are already failing and I don't know the responses we'll get once security is actually added to HTTP.  I therefore propose we push out fleshing out those tests until we actually can support this funcationality.

andytill:
OK, do we need to maintain that matrix of tests?

javajolt:
Jira issue tracking HTTP security: https://bashoeng.atlassian.net/browse/RTS-1258

macintux:
:+1:

PR #1109, opened 2016-07-22
SQL `GROUP BY` RTS-732

andytill:
Tests for SQL `GROUP BY` functionality and upgrade.

### DIFF
--- a/tests/ts_cluster_capabilities_SUITE.erl
+++ b/tests/ts_cluster_capabilities_SUITE.erl
@@ -47,6 +47,7 @@ end_per_group(_GroupName, _Config) ->
 init_per_testcase(_TestCase, Config) ->
     %% tear down the whole cluster before every test
     rtdev:setup_harness('_', '_'),
+    ct:pal("TEST CASE ~p", [_TestCase]),
     Config.
 
 end_per_testcase(_TestCase, _Config) ->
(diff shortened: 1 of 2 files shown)


tests/ts_cluster_capabilities_SUITE.erl
```diff
@@ -0,0 +1,222 @@
...
+groups() ->
+    [].
+
+all() -> 
+    rt:grep_test_functions(?MODULE).
+
+-define(TS_VERSION_CURRENT, "current").
+
+%% git checkout riak_ts-1.3.1 && make locked-deps
+-define(TS_VERSION_1_3, "ts_1.3.1").
```
javajolt: Do we really want to hard-code this? `riak_test` has the notion of `previous` for this reason.  It also has capabilities to set up an old cluster.

andytill: 'fraid so, in the next release previous will be 1.4 and current will be 1.5 so this test won't mean the same thing.

javajolt: Ugh.  That's a huge testing burden because we'll have to keep that devrel version around and distributed everywhere until this test dies.  There is something very 1.3.1-specific?  If so, let's rename the test to reflect that.

andytill: It is 1.3.1 specific, this version doesn't have the select record capability.  At some point, some of these tests will die because there is no chance of having a mixed version cluster with 1.3.1 in.  Happy to rename to whatever makes it easier.

javajolt: What about this: Could `previous` be 1.3.1?  How flexible is this? Would it run potentially on 1.4 -> 1.5 upgrade, too?

andytill: I don't understand what you are proposing.

javajolt:
- `ts_cluster_capabilities_SUITE` passes if specific devrel name is used
- `ts_cluster_group_by_SUITE`

gordonguthrie:
+1 524cce3

PR #1110, opened 2016-07-26
RTS-1051: Add ts_simple_show_tables

javajolt:
Add a simple test to create a handful of tables then make sure that SHOW TABLES does the expected thing


paegun:
+1 4e979f9

paegun:
+1 57481f1

PR #1113, opened 2016-08-01
RTS-1271: Fix DESCRIBE and add SHOW TABLES tests

javajolt:
Add  a test for SHOW TABLES and fix DESCRIBE output


paegun:
+1 1e12b81

PR #1114, opened 2016-08-02
r-e-c now expects good unicode, so use "~ts" in Queries to accommodate

hmmr:
This fixes regressions in ts_simple_insert (specifically, by constraining generated varchars to `$a..$z`), and in ts_cluster_unicode.


macintux:
Good find, thanks. Will review soon.

macintux:
Tests pass, +1

PR #1117, opened 2016-08-03
Rejig of updown test suite to add group_by tests

gordonguthrie:


tests/ts_cluster_updowngrade_group_by_SUITE.erl
```diff
@@ -0,0 +1,167 @@
...
+ 
+sorted_assert(String, {ok, {ECols, Exp}}, {ok, {GCols, Got}}) ->
+    Exp2 = lists:sort(Exp),
+    Got2 = lists:sort(Got),
+    ts_util:assert_float(String, {ECols, Exp2}, {GCols, Got2});
+sorted_assert(String, Exp, Got) ->
+    ok = log_error(String ++ " banjo", Exp, Got),
+    fail.
+
+plain_assert(_String, Exp, Exp) ->
```
hmmr: I think you should still use `assert_float` here as well?

macintux: That does seem to be desirable. One of the tests uses `avg` and thus includes floats in the expected results.

tests/ts_updown_util.erl
```diff
@@ -341,33 +341,35 @@ fmt(F, A) ->
...
     DDL = fmt(DDLFmt, [Table]),
     Client1 = rt:pbc(TableNode),
     case riakc_ts:'query'(Client1, DDL) of
         Exp ->
             ok = wait_until_active_table(Client1, Table, 5),
             ct:log("Table ~p created on ~p", [Table, TableNode]),
             pass;
-        Error ->
-            ct:log("Failed to create table ~p: (~s)", [Table, Error]),
+        {error, {_No, Error}} ->
```
macintux: Are we certain the error messages from query are always going to be in the form of a line number tuple? The specs in `riak-erlang-client` do not "enforce" that.

macintux:
With the minor exception of the concerns noted, this looks good.

javajolt:
8 hours later the tests finished

javajolt:
:+1: 94ae0c3

PR #1118, opened 2016-08-04
RTS-1050: Add tests for EXPLAIN command

javajolt:
I fear that the `riak_shell` test may fail because I think it truncates the output based on the size of the window.  Maybe I could standardize to 80 columns? Not sure how GiddyUp runs it...

### DIFF
--- a/priv/riak_shell/riak_shell_regression1.log
+++ b/priv/riak_shell/riak_shell_regression1.log
@@ -24,6 +24,19 @@
 |GeoCheckin|
 +----------+
 "}}.
+{{command, "EXPLAIN SELECT * FROM GeoCheckin WHERE myfamily = 'family1' AND myseries = 'series1' AND time >= 2 AND time <= 7000000 AND weather='fair';\n"}, {result, "+-----+-------------------------------------+------------+-------------------------------------+-----------+----------+
+|Subqu|        Range Scan Start Key         |Is Start Inc|         Range Scan End Key          |Is End Incl|  Filter  |
++-----+-------------------------------------+------------+-------------------------------------+-----------+----------+
+|  1  |myfamily = 'family1', myseries = 'ser|   false    |myfamily = 'family1', myseries = 'ser|   false   |weather = |
+|  2  |myfamily = 'family1', myseries = 'ser|   false    |myfamily = 'family1', myseries = 'ser|   false   |weather = |
+|  3  |myfamily = 'family1', myseries = 'ser|   false    |myfamily = 'family1', myseries = 'ser|   false   |weather = |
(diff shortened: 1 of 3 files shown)


tests/ts_util.erl
```diff
@@ -315,6 +315,7 @@ get_valid_big_data(N) ->
         N + 0.1,
         get_bool(X),
         N + 100000,
+        get_varchar(),
```
andytill: Will this break inserts and result set assertions for other tests using this shared table def?

tests/ts_simple_explain_query.erl
```diff
@@ -0,0 +1,64 @@
...
+
+-behavior(riak_test).
+
+-export([confirm/0]).
+
+-include_lib("eunit/include/eunit.hrl").
+
+%% Test basic table description
+
+confirm() ->
```
andytill: Can this be in ct style, so more tests can be added later on?

andytill:
There needs to be a test for security permissions one where security is enabled and the use has permission, and one where they do not and the request is denied.

priv/riak_shell/riak_shell_regression1.log
```diff
@@ -24,6 +24,20 @@
 |GeoCheckin|
 +----------+
 "}}.
+{{command, "EXPLAIN SELECT * FROM GeoCheckin WHERE myfamily = 'family1' AND myseries = 'series1' AND time >= 2 AND time <= 7000000 AND weather='fair';\n"}, {result, "+-----+-------------------------------------+------------+-------------------------------------+-----------+----------+
+|Subqu|        Range Scan Start Key         |Is Start Inc|         Range Scan End Key          |Is End Incl|  Filter  |
++-----+-------------------------------------+------------+-------------------------------------+-----------+----------+
+|  1  |myfamily = 'family1', myseries = 'ser|   false    |myfamily = 'family1', myseries = 'ser|   false   |weather = |
```
andytill: Weather should be 'fair' in the filter.

tests/ts_util.erl
javajolt: I added this in response to my schema change below. The 3 other tests I found using the big schema passed with the change.

priv/riak_shell/riak_shell_regression1.log
paegun: The columns are truncated, so the "Filter" value isn't visible. This is also the case for the "Range Scan Start Key" column and other column displays.

paegun:
+1 ebea138

priv/riak_shell/riak_shell_regression1.log
andytill: Sounds like it needs some issues fixed before it can be used for testing.

paegun:
+1 2d905b2

PR #1119, opened 2016-08-05
Added support for specifying an upgrade callback function

fadushin:
This function be passed to the rt:upgrade function and which gets called after the from node is stopped and upgraded, and before the to node is started.  This allows tests to perform any additional steps as part of the upgrade.  The paths of the old and new data directories are passed elements of a prop list, which is passed into the callback as its sole argument.


fadushin:
Wrong branch.  killing.

PR #1120, opened 2016-08-05
Added support for specifying an upgrade callback function, which can …

fadushin:
…be passed to the rt:upgrade function and which gets called after the from node is stopped and upgraded, and before the to node is started.  This allows tests to perform any additional steps as part of the upgrade.  The paths of the old and new data directories are passed elements of a prop list, which is passed into the callback as its sole argument.

### DIFF
--- a/src/rtssh.erl
+++ b/src/rtssh.erl
@@ -250,10 +250,12 @@ stop(Node) ->
     run_riak(Node, "stop"),
     ok.
 
-upgrade(Node, NewVersion) ->
-    upgrade(Node, NewVersion, same).
+upgrade(Node, NewVersion, UpgradeCallback) when is_function(UpgradeCallback) ->
+    upgrade(Node, NewVersion, same, UpgradeCallback).
 
-upgrade(Node, NewVersion, Config) ->
+%% upgrade callback unsupported for this driver until there is a need.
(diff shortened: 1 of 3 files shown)


src/rtssh.erl
```diff
@@ -250,10 +250,13 @@ stop(Node) ->
...
-upgrade(Node, NewVersion) ->
-    upgrade(Node, NewVersion, same).
+upgrade(Node, NewVersion, UpgradeCallback) when is_function(UpgradeCallback) ->
+    upgrade(Node, NewVersion, same, UpgradeCallback).
 
-upgrade(Node, NewVersion, Config) ->
+%% upgrade callback unsupported for this driver until there is a need.
+%% c.f., rtdev:upgrade/4
+upgrade(Node, NewVersion, Config, _UpgradeCallback) ->
+    lager:warning("Upgrade callback not supported in the rtssh riak test harness!"),
```
nickelization: So the trouble I see with this is that we're going to see this warning printed *every* time an upgrade is run using rtssh, regardless of whether the test specifies an upgrade callback or not. (We will still get `fun no_op/1` passed in here.)

nickelization:
+1 45a36c1

PR #1121, opened 2016-08-10
Compensate for integer vs string (Erlang client vs riak-shell)

macintux:
Erlang client: integer timestamps
riak-shell: ISO 8601 strings

Bridge the difference for testing.


javajolt:
:+1: 3ac8102

PR #1122, opened 2016-08-11
Upload debug logs to Giddyup

jvoegele:
This can be disabled by setting the `giddyup_post_debug_logs` config value to `false` (it is `true` by default).

### DIFF
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -297,25 +297,44 @@ run_test(Test, Outdir, TestMetaData, Report, HarnessArgs, NumTests) ->
         1 -> keep_them_up;
         _ -> rt:teardown()
     end,
-    CoverageFile = rt_cover:maybe_export_coverage(Test, CoverDir, erlang:phash2(TestMetaData)),
+    CoverageFile = rt_cover:maybe_export_coverage(Test,
+                                                  CoverDir,
+                                                  erlang:phash2(TestMetaData)),
     case Report of
         undefined -> ok;
         _ ->
-            {value, {log, L}, TestResult} = lists:keytake(log, 1, SingleTestResult),
+            {value, {log, L}, TestResult} =
+                lists:keytake(log, 1, SingleTestResult),
             case giddyup:post_result(TestResult) of
                 error -> woops;
                 {ok, Base} ->
                     %% Now push up the artifacts, starting with the test log
                     giddyup:post_artifact(Base, {"riak_test.log", L}),
-                    [ giddyup:post_artifact(Base, File) || File <- rt:get_node_logs() ],
-                    [giddyup:post_artifact(Base, {filename:basename(CoverageFile) ++ ".gz",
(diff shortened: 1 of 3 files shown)


src/riak_test_escript.erl
```diff
@@ -297,26 +297,45 @@ run_test(Test, Outdir, TestMetaData, Report, HarnessArgs, NumTests) ->
         1 -> keep_them_up;
         _ -> rt:teardown()
     end,
-    CoverageFile = rt_cover:maybe_export_coverage(Test, CoverDir, erlang:phash2(TestMetaData)),
```
nickelization: No need to break up these long lines - last I heard, anything below 100 characters is fair game. (I know some people will say 80, but I believe the consensus was that the 80 char limit is a holdover from old line printers and 80x25 terminals, and is unnecessarily short for today's monitors.)

src/riak_test_escript.erl
```diff
@@ -297,26 +297,45 @@ run_test(Test, Outdir, TestMetaData, Report, HarnessArgs, NumTests) ->
...
         _ ->
-            {value, {log, L}, TestResult} = lists:keytake(log, 1, SingleTestResult),
+            {value, {log, L}, TestResult} =
+                lists:keytake(log, 1, SingleTestResult),
             case giddyup:post_result(TestResult) of
                 error -> woops;
                 {ok, Base} ->
                     %% Now push up the artifacts, starting with the test log
                     giddyup:post_artifact(Base, {"riak_test.log", L}),
-                    [ giddyup:post_artifact(Base, File) || File <- rt:get_node_logs() ],
```
nickelization: Also, if you do find lines that are still too long, could you chop them up in a separate commit next time please? When it's all in one commit, it makes it tricky to identify what code has actually changed vs. what has just been reformatted.

src/rtdev.erl
```diff
@@ -765,6 +778,26 @@ get_node_logs() ->
...
+    rt:pmap(fun(Node) ->
+                    get_node_debug_logs(Node)
+            end,
+            NodeMap).
+
+get_node_debug_logs({_Node, NodeNum}) ->
+    Path = relpath(node_version(NodeNum)),
+    Args = ["--logs"],
+    Cmd = riak_debug_cmd(Path, NodeNum, Args),
+    file:delete(?DEBUG_LOG_FILE(NodeNum)),
```
nickelization: Just by inspection, this code feels slightly odd. I'm guessing I'm just missing something, but it looks like we delete the log file before we actually spawn the command to upload it. Even if everything works fine here, I'd like to see a comment or two explaining more clearly what's actually going on.

jvoegele: I refactored to use a (documented) function to delete the existing log file.

src/riak_test_escript.erl
jvoegele: Yeah, I'd thought that 80 was our agreed upon convention, but I'm more than willing to use 100 instead.

src/riak_test_escript.erl
jvoegele: Duly noted.

nickelization:
:+1: cfccfdd

PR #1124, opened 2016-08-12
Create riak.conf under 1.3 when downgrading RTS-1302

andytill:
TS 1.3 cannot naturally read configuration created under the latest 1.4, after this test was written.  The downgrade test is changed so that Node A starts at 1.3, then upgrades to 1.4, then downgrades to 1.3 so that it's original riak.conf is created under 1.3.

In the case that the user starts off with a 1.4 node then wants to downgrade to 1.3, manual changes to the property names in riak.conf must be made, the error is quite clear what they should be though, and riak will not start until it finds the correct property names.

Also used the capability atoms instead of integers for the select capability.

### DIFF
--- a/tests/ts_cluster_capabilities_SUITE.erl
+++ b/tests/ts_cluster_capabilities_SUITE.erl
@@ -148,7 +148,7 @@ sql_select_upgrade_a_node_from_1_3_test(_) ->
     ok = rt:wait_until_ring_converged([Node_A,Node_B,Node_C]),
     ok = rt:upgrade(Node_A, ?TS_VERSION_CURRENT),
     ok = rt:wait_until_ring_converged([Node_A,Node_B,Node_C]),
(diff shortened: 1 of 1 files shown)


hmmr:
I'm fine with the PR, but can you clarify on how this test is going to work on version combinations other than 1.3 and 1.4?

andytill:
@hmmr  There are no other version combinations to test right now, they will need their own tests, or be added to the upgrade downgrade suite.

hmmr:
@andytill which means, at some point the module will need to be renamed (or modified) to make it clear it only applies to 1.3 vs 1.4, right? My argument here is that it should be done now.

andytill:
Once we go past the point where direct upgrade and downgrade from 1.3 is not supported, those tests can be deleted.

hmmr:
+1 43b71a0

PR #868, opened 2015-09-17
bdp spark pi submission test

korry8911:

Issue #184, opened 2013-01-28
Loaded Upgrade Search issues aka The Mister Zero bug


joedevivo:
There is a known issue with riak_search returning bad search results while the cluster is in the middle of an upgrade. This issue makes itself known with a riak_test failure like this 

```
{{assertEqual_failed,[{module,loaded_upgrade},
                      {line,340},
                      {expression,"Actual"},
                      {expected,[{"YALSP",3},
                                 {"ZiaSun",1},
                                 {"headaches",4},
                                 {"mister",0},
                                 {"mx.example.net",187},
                                 {"prohibiting",5}]},
                      {value,[{"YALSP",3},{"ZiaSun",1},{"mister",0}]}]},
```

and gets it's name from the last line.

Digging in to why this happens we noticed some issues in using basho_bench to verify results, having to do is that basho bench is very black and white in terms of errors, whereas riak test in this scenario needs more of a grayscale (grayskull?) system. We also noticed that search checks will fail (in this known case) but riak_test ignores the failure. 

Work is already in progress to apply load to a cluster for loaded upgrade in a way that allows riak_test to make decisions and report on a complete upgrade.

PR #1126, opened 2016-08-15
RIAK-2583 Async Job Switches

tburghart:
Adds system-level tests for:
* RPC enable/disable management.
* Operation functionality through HTTP and protobuf interfaces.

Test Modules:
* verify_job_enable_ac
* verify_job_enable_rc
* verify_job_switch_defaults

Also adds `rt:random_uniform/0` and `rt:random_uniform/1` wrappers around `random:uniform/n` that ensure the process's PRNG is seeded _once_ with quality entropy.

Prerequisites:
- [x] [riak_core PR #851 (RIAK-2668)](https://github.com/basho/riak_core/pull/851)
- [x] [riak_kv PR #1459](https://github.com/basho/riak_kv/pull/1459)
- [x] [riak_search PR #184 (RIAK-1872) (RIAK-1920) (RIAK-1944) (RIAK-2349)](https://github.com/basho/riak_search/pull/184)
- [x] [yokozuna PR #671 (RIAK-2187)](https://github.com/basho/yokozuna/pull/671)
- [x] [riak PR #868](https://github.com/basho/riak/pull/868)
- [x] [riak_ee PR #405 (RIAK-1410)](https://github.com/basho/riak_ee/pull/405)

### DIFF
--- a/tests/job_enable_common.erl
+++ b/tests/job_enable_common.erl
@@ -1,0 +1,362 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2016 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(job_enable_common).
+
+% Public API for use by other tests
+-export([
+    bin_bucket/1,
+    bin_key/1,
+    bin_val/1,
+    close_client/1,
+    enabled_string/1,
+    index_2i/0,
+    index_yz/0,
+    load_data/1,
+    num_buckets/0, num_buckets/1,
+    num_keys/0, num_keys/1,
+    open_client/2,
+    populated_bucket/0,
+    setup_cluster/1,
+    test_buckets/0,
+    test_keys/0,
+    test_nums/0,
+    test_operation/4,
+    test_vals/0,
+    undefined_bucket/0
+]).
+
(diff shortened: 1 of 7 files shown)


tests/job_enable_common.erl
```diff
@@ -0,0 +1,572 @@
...
+%% Test API
+%% ===================================================================
+
+enabled_string(true) ->
+    "enabled";
+enabled_string(false) ->
+    "disabled".
+
+bin_bucket(Num) ->
+    erlang:list_to_binary(["Bucket_", erlang:integer_to_list(Num)]).
```
nickelization: Not worth changing now, but just for the future, you really don't have to qualify functions like `list_to_binary` with `erlang:` - we use these builtin type conversion functions all over the place without the extra verbosity of specifying the `erlang` module. So, there really shouldn't be any ambiguity caused by following the convention of leaving off the module, and the code is marginally easier to read without the extra verbosity IMHO.

tests/job_enable_common.erl
```diff
@@ -0,0 +1,572 @@
...
+                    ?assertEqual({error, ?ERRMSG_BIN(Class)}, Result);
+                http ->
+                    ?assertMatch({error, {"403", _}}, Result)
+            end
+    end;
+
+test_operation(Node, ?TOKEN_SEC_INDEX = Class, Enabled, pbc = ClientType) ->
+    Bucket = populated_bucket(),
+    Index = index_2i(),
+    Num = random:uniform(num_keys()),
```
nickelization: I'm not sure if this ultimately really matters or not, but it looks like we're never seeding the `random` module. This means it will automatically call `random:seed/0` the first time we call `random:uniform`, which uses a fixed value for the seed, which means we'll get the same sequence of "random" numbers every time we run a test.

nickelization:
@tburghart I give this a conditional +1 - if you're not concerned about the deterministic random numbers, I say go ahead and merge this. Otherwise let me know once you have any new changes in place and I'll re-review.

tests/job_enable_common.erl
tburghart: ha ha ha - I won't add my rant about the evils of unqualified names here, but it's quite deliberate ;)

nickelization: Alright well...since we have zero established coding standards I suppose we'll just have to agree to disagree ;). If this were code in an existing module that used the unqualified style, I'd say we shouldn't introduce style inconsistencies, but since it's a new module I guess at least we're being consistent.

nickelization:
+1 6aa340a

PR #1128, opened 2016-08-17
Downgrade riak.conf to 1.3 RTS-1303

andytill:
When downgrading a node to 1.3, replace the TS properties in riak.conf with the old names.

### DIFF
--- a/src/rtdev.erl
+++ b/src/rtdev.erl
@@ -155,6 +155,10 @@ upgrade(Node, NewVersion, Config) ->
     rt_config:set(rt_versions, VersionMap),
     case Config of
         same -> ok;
+        _ when is_function(Config) ->
+            RiakConfPath =
+                io_lib:format("~s/dev/dev~b/etc/riak.conf", [NewPath, N]),
+            Config(RiakConfPath);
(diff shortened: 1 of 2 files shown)


macintux:
This needs to take into account `riak_kv.query.timeseries.timeout` as well.

We also need a solution for non-TS tests that suffer from the same problem, so if you have any ideas on the simplest fix for tests that rely on `rt:upgrade`, that'd be terrif.

E.g., https://github.com/basho/riak_test/blob/develop/tests/replication_object_reformat.erl#L83-L85

andytill:
The timeout property cannot be handled like the others, have a look at the comment code here https://github.com/basho/riak_test/pull/1128/files#diff-4331a830e929df626095895f5238cb7fR308

In the other tests I think we should just export this function and pass it in as the `rt:upgrade/3` `Config` argument.

macintux:
We discussed this yesterday. Customers will only run into this issue if they install a new version from scratch and attempt a downgrade, which seems unlikely.

An aborted rolling upgrade scenario could do it, however.

-John

javajolt:
Seems to be a duplicate now of #1129, right?

PR #1130, opened 2016-08-17
Update client libraries to use develop-2.2

nickelization:
This pulls in the latest client code, as well as newer riak_pb stuff
which is needed for the latest develop-2.2 Riak code.


tburghart:
👍 30b0c01

PR #1131, opened 2016-08-18
Allow devrels to be looked up by name instead of alias

javajolt:
As long as the legacy version riak_ts-1.3.1 is installed, i.e., is in current, previous or legacy (or specially installed) this test should now pass.  Previously `ts_cluster_capabilities_SUITE` assumed a specially named devrel install of `riak_ts-1.3.1`.  This works better for GiddyUp and once the legacy version is no longer supported this test will start failing and should be updated or removed from the test suite.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -65,6 +65,7 @@
          down/2,
          enable_search_hook/2,
          expect_in_log/2,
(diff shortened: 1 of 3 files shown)


paegun:
+1 df5fdcc

PR #1133, opened 2016-08-19
Stop more nodes on riak_shell_test_disconnecting

javajolt:
Try stopping another node to make the test consistent


paegun:
+1 09bc84d
Makes sense to make this more deterministic. One node up should surely lead to that node being the one that reconnect connects to.

PR #1134, opened 2016-08-29
Add back call to stop nodes before deleting index directories.

JeetKunDoug:
Somehow managed to pass on 2.0.7, but shouldn't have (and doesn't in 2.2.0).


jvoegele:
+1 70ce356

fadushin:
+1 70ce356

PR #1135, opened 2016-08-30
Wait for CRDT search results, rather than failing after the first attempt.

JeetKunDoug:
Simple rt_wait_until wrapper around the search validation code so we can allow for multiple attempts.


fadushin:
+1 61c1ea8

PR #1137, opened 2016-08-30
Changed extractor test to use bad_entry stat

JeetKunDoug:
Stats for index_fail and bad_entry have been split and the double-counting of index_fail stats has been fixed. Update the test to look at the correct stat and remove the compensation for double-counting failures.


JeetKunDoug:
+1 d4bc7f4

JeetKunDoug:
Will pass currently on dr-solrq-per-partition-index branch... will fail until that is merged.

PR #1141, opened 2016-09-07
Reorganize ts_util into new modules, more explicit function calls

macintux:
`ts_util.erl` has grown into a bit of a mess, exemplified by `ts_util:ts_query` which, despite its name, does the following tasks:
- Create a bucket type
- Activate the bucket type
- Populate the cluster with data
- Query the data

This PR creates 3 modules to do the same tasks as `ts_util`: `ts_setup`, `ts_data`, and `ts_ops`. It also converts a simple test.

Ideally we'd migrate all tests away from `ts_util`, and perhaps convert those that have largely abandoned it because the functions are too opaque and not nearly granular enough for complex tests.

### DIFF
--- a/tests/ts_data.erl
+++ b/tests/ts_data.erl
@@ -1,0 +1,438 @@
+%% -*- Mode: Erlang -*-
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2015, 2016 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%% @doc A util module for riak_ts tests
+
+-module(ts_data).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-export([get_ddl/0, get_ddl/1, get_ddl/2,
+         assert/3,
+         assert_error_regex/3,
+         assert_float/3,
+         assert_row_sets/2,
+         exclusive_result_from_data/3,
+         flat_format/2,
+         get_bool/1,
+         get_cols/0, get_cols/1,
+         get_data/1,
+         get_default_bucket/0,
+         get_float/0,
+         get_integer/0,
+         get_invalid_obj/0,
+         get_invalid_qry/1,
+         get_long_obj/0,
+         get_map/1,
+         get_optional/2,
+         get_short_obj/0,
+         get_string/1,
+         get_timestamp/0,
+         get_valid_aggregation_data/1,
+         get_valid_aggregation_data_not_null/1,
+         get_valid_big_data/1,
+         get_valid_obj/0,
+         get_valid_qry/0, get_valid_qry/1,
+         get_valid_qry/2, get_valid_qry/3,
+         get_valid_qry_spanning_quanta/0,
+         get_valid_select_data/0, get_valid_select_data/1,
+         get_valid_select_data_spanning_quanta/0,
+         get_varchar/0,
+         remove_last/1,
+         results/1
+        ]).
+
+-define(MAXVARCHARLEN,       16).
+-define(MAXTIMESTAMP,        trunc(math:pow(2, 63))).
+-define(MAXFLOAT,            math:pow(2, 63)).
+
+%% This is also the name of the table
+get_default_bucket() ->
+    "GeoCheckin".
+
+get_valid_qry() ->
+    get_valid_qry(get_default_bucket()).
+
+get_valid_qry(Table) ->
+    "select * from " ++ Table ++ " Where time > 1 and time < 10 and myfamily = 'family1' and myseries ='seriesX'".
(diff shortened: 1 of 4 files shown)


tests/ts_setup.erl
```diff
@@ -0,0 +1,100 @@
...
+-spec stop_a_node(list(node())) -> ok.
+stop_a_node(Nodes) ->
+    stop_a_node(2, Nodes).
+
+-spec stop_a_node(pos_integer(), list(node())) -> ok.
+stop_a_node(Which, Nodes) ->
+    ok = rt:stop(lists:nth(Which, Nodes)).
+
+-spec create_bucket_type(list(node()), string(), string()) -> ok.
+create_bucket_type(Nodes, DDL, Table) ->
```
hmmr, line 66: The two-step procedure to create a TS table, involving riak-admin, dates back to the time when we didn't have a working CREATE TABLE query. While it can be seen as making sense (if you squint) to, for instance, check that `riak-admin`'s function remains not broken, creating a table with a single query is a cleaner and perhaps faster means to our testing ends. I would suggest to ~~deprecate~~ delete the `create_bucket_type` and `activate_bucket_type` functions and replace them at call sites with `{ok, {[], []}} = riakc_ts:query(Client, DDL)`.

tests/ts_data.erl
```diff
@@ -0,0 +1,438 @@
...
+
+exclusive_result_from_data(Data, Start, Finish) when is_integer(Start)   andalso
+                                                     is_integer(Finish)  andalso
+                                                     Start  > 0          andalso
+                                                     Finish > 0          andalso
+                                                     Finish > Start ->
+    lists:sublist(Data, Start, Finish - Start + 1).
+
+remove_last(Data) ->
+    lists:reverse(tl(lists:reverse(Data))).
```
hmmr: By way of suggestion: `lists:sublist(Data, length(Data) - 1)`.

tests/ts_simple_select.erl
```diff
@@ -27,14 +27,18 @@
...
-            ts_util:cluster_and_connect(single), TestType, DDL, Data, Qry),
+        {ok, {ts_data:get_cols(small),
+         ts_data:exclusive_result_from_data(Data, 2, 9)}},
+
+    Cluster = ts_setup:start_cluster(1),
+    ts_setup:create_bucket_type(Cluster, DDL, Table),
+    ts_setup:activate_bucket_type(Cluster, Table),
+    ts_ops:put(Cluster, Table, Data),
+    Got = ts_ops:query(Cluster, Qry),
     ?assertEqual(Expected, Got),
```
andytill, line 43: It is still not possible to work out what this test without doing a lot of lookup.

The test gets some kind of DDL, the some kind of valid data, then a valid query, then some columns and results, but we need to start at 2 for some reason.

I know this was around way before this change but cleaning up the canned test data perpetuates a very bad pattern of not being able to understand what a test does by looking at it.  Why not leave the data as it is, but use the actual SQL CREATE TABLE statements, data and query in the test itself?

tests/ts_data.erl
macintux: Pushed.

tests/ts_setup.erl
macintux: I'm disinclined to clean that up with this PR in interests of prioritizing other activities. Any objections?

andytill (changes requested):

hmmr:
It's a good idea, but the PR itself is half-done -- exactly for the reason @macintux gives in the PR description. It would not be complete without a comprehensive, global replace of calls to migrated function, and the removal of them from `ts_util`, I'm afraid.

tests/ts_simple_select.erl
hazen: @andytill Code maintenance is a huge issue. If each test has a static schema and data, then each test individually needs to be updated when a sweeping change happens, rather than updating this in one library.  Sadly we've had several big changes over the life of TS.  Many of the functions in the utility library are simply setting up a cluster into an interesting test state, but the details are less important.  Much like building a cluster.  I would agree that the interesting features of what are being tested should be in the test, however.

andytill: If you add up to the time we spent converting tests to use row containers from lists to tuples, I doubt it was a lot, and now it is over. If I added up the amount of time I spend trying to work out what some of our tests are doing it would be much longer! And I pay it every time I need to read the tests.  Tests are read heavy code, we should optimise them to be read in isolation as opposed to a part of the bigger framework.  This is the opposite of how I would write most production code of course.

This is one of those sweeping changes, but it doesn't affect those tests that are isolated. That don't use the utility functions and have shared data.  Re-usability also has costs.

macintux: I readily admit that this is more granular but almost as opaque as the current testing library.

I'm fine with a preference for a better testing interface, but I'd like to see a concrete example so we can evaluate how much effort it would be to create and to convert the tests to use it.

andytill: `ts_cluster_keys_SUITE.erl` is roughly where we should be heading.  The data is in or near the test, it doesn't use util functions that do anything special and it is common test so lots of small, individual tests.

It could do with some improvements. We can make more use of groups where several tests run over one table, by creating the in `init_per_group/2` the whole group will fail if the table can't be created, but not affect other groups like it does it the keys suite.  I've also made use of groups to test with PB and TTB with the same tests.

Converting the tests doesn't have to be a big bang thing, if plan to convert a few each release, or when tackling a relevant feature we'll soon get through them!

https://github.com/basho/riak_test/blob/riak_ts-develop/tests/ts_cluster_keys_SUITE.erl

hazen:
Changes included in https://github.com/basho/riak_test/pull/1187.  Probably just want to close and open a new one for additional work.

andytill:
Merged in https://github.com/basho/riak_test/pull/1187

PR #1143, opened 2016-09-07
Fix riak_kv_get_fsm intercept used by overload

nickelization:
The get FSM normally uses `start/4` instead of `start_link/4` now, which
caused the overload test to break. Fixed the intercept name and all is
working again now.


JeetKunDoug:
+1 f785ee7

PR #1144, opened 2016-09-08
Disable use of job enable flags for this test

nickelization:
If we leave it up to the Cuttlefish defaults, we end up with search 1.0 being disabled, which causes this test to fail. Setting the config option to `undefined` causes us to revert to the old behavior of everything being enabled.


javajolt:
Testing...

javajolt:
:+1: beb2083

PR #1149, opened 2016-09-12
Modified rtdev:get_node_debug_logs to return filename and file handle

fadushin:
so that the debuglog archive can be uploaded.


jvoegele:
+1 396a6cf

PR #1150, opened 2016-09-13
RTS-1391: Test EXPLAIN with PBC, not TTB, which expects binaries

javajolt:
Requires https://github.com/basho/riak_kv/pull/1478


andytill:
+1 74be90e

PR #1151, opened 2016-09-14
Add copy_conf function and fix verify_riak_object_reformat test

nickelization:
The new `copy_conf` utility function copies config from one devrel to another. This is particularly useful when the first thing a test does is downgrade to a previous release which doesn't support some new config option that's been recently added. By copying the old config into the new devrel before we start up the nodes, we can automatically avoid specifying any new, incompatible config keys for that particular test.

This PR also uses `copy_conf` to fix the `verify_riak_object_reformat` test, which had been failing during the initial downgrade due to unrecognized cuttlefish keys.

### DIFF
--- a/tests/verify_riak_object_reformat.erl
+++ b/tests/verify_riak_object_reformat.erl
@@ -37,7 +37,7 @@ confirm() ->
 
     %% Use previous version's riak.conf so that when we
     %% downgrade we don't crash on unknown config keys:
-    rtdev:copy_conf(?N, previous, current),
+    rt:copy_conf(?N, previous, current),
 
     Nodes = [Node1|_] = rt:build_cluster(?N),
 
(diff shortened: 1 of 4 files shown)


tests/verify_riak_object_reformat.erl
```diff
@@ -34,6 +34,11 @@ confirm() ->
     rt:update_app_config(all, [{riak_kv, [{object_format, v1}]}]),
     TestMetaData = riak_test_runner:metadata(),
     DowngradeVsn = proplists:get_value(upgrade_version, TestMetaData, previous),
+
+    %% Use previous version's riak.conf so that when we
+    %% downgrade we don't crash on unknown config keys:
+    rtdev:copy_conf(?N, previous, current),
```
fadushin: This should be routed through rt, not called directly.

(Another alternative would be to use the upgrade callback mechanism, but this is fine, too)

javajolt: Good catch, @fadushin

nickelization: Ah right, makes sense. Thanks @fadushin, will fix and add another commit shortly.

fadushin:
+1 8e5505d

PR #1152, opened 2016-09-14
new module to check query buffer operations

hmmr:
RTS-733, RTS-870, RTS-1006

Depends on https://github.com/basho/riak_kv/pull/1479, which see.


tests/ts_cluster_coverage.erl
```diff
@@ -1,6 +1,6 @@
 % -------------------------------------------------------------------
```
gordonguthrie, line 1: Running this test thows crash reports

```
2016-10-16 08:52:05 =ERROR REPORT====
Error in process <0.155.0> on node 'riak_test@127.0.0.1' with exit value: {{nocatch,got_right_results_from_wrong_vnode},[{ts_cluster_coverage,make_noise_if_results_not_empty,5,[{file,"tests/ts_cluster_coverage.erl"},{line,183}]},{ts_cluster_coverage,'-test_quanta_range/5-fun-1-',3,[{file...
```

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -0,0 +1,314 @@
...
+            F5 <- ?ORDBY_COLS,
+            all_different([F1, F2, F3, F4, F5]),
+            not_all_undefined([F1, F2, F3, F4, F5])],
+    OrderByBattery2 =
+        lists:map(
+          fun(FF) ->
+                  [F || F <- FF, F /= undefined]
+          end,
+          OrderByBattery1),
+    lists:foreach(
```
gordonguthrie, line 114: Anonymous functions inside `list:foreach` are hard to decipher, doubling it makes it worse and makes the risk of unspotted unintended capture very, very high

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -0,0 +1,314 @@
...
+    {ok, {_Cols, Returned}} =
+        riakc_ts:query(C, Query, [], undefined, Options),
+    OrderBy = proplists:get_value(order_by, Clauses),
+    Limit   = proplists:get_value(limit, Clauses),
+    Offset  = proplists:get_value(offset, Clauses),
+    %% emulate the SELECT on original data,
+    %% including sorting according to ORDER BY
+    %% uncomment these log entries to inspect the data visually (else
+    %% it's just going to blow up the log into tens of megs):
+    %% ct:log("Query: \"~s\"", [Query]),
```
gordonguthrie, line 282: `ct:pal` is probably better here

gordonguthrie (commented):
We need to review the test suites again in light of getting a list of all error conditions. Also the test suite needs to extend from simple to cluster and tests need to be added to the upgrade/downgrade ladder.

Once again we need to think about writing the EQC SQL generator tests...

hmmr:
The commits above, along with the new module ts_cluster_updowngrade_order_by_SUITE, contain these fixes to the TS up/downgrade facilities:
1. Correctly set up and start the previous client slave node. Add `{use_previous_client, true}` to the proplist in your `Config`, which needs to be returned by `make_initial_config(Config)` to take effect (  4b61877).
2. Add a pair of hooks, to be called before a transitioned node is started, with the path to its riak.conf. Useful when the newer version introduces new keys which will need to be removed from the same riak.conf before the older riak version can use it. This was previously hardcoded for the case of 1.3 vs 1.4, in ts_cluster_updowngrade_group_by_SUITE, and is now generalized (229f809, 7707de0).
3. After starting a transitioned node, we need to give it some extra time for the changed capabilities to percolate in the cluster (e188bfa).

tests/ts_simple_query_buffers_SUITE.erl
hmmr: This is carefully hand-crafted to generate a those test queries. It's definitely not the most readable code. To help understand its purpose, I'm specifically advising to uncomment the `ct:log` entries in `check_sorted` printing each query as we go (but then comment it again because of the huge volumes of logs it would generate).

gordonguthrie:
+1 eff8df9

hmmr:
@borshop merge

PR #1153, opened 2016-09-14
Improve join cluster retry logic

nickelization:
There is a weird issue we've repeatedly seen during testing where we end up planning and committing a join during cluster setup, only to get back `{error, nothing_planned}`. The current code ends up ignoring and continuing on if it happens more than once, yet sometimes it seems the plan really didn't commit for whatever reason. I'm not sure if retrying more times will fix the issue...but it should at least shed some light on whether we're just seeing the problem coincidentally occur twice in a row, or whether the cluster is getting stuck in some kind of bad state where no amount of retrying will get us unstuck.

We should still address this issue on the Riak side eventually, but it's been present for multiple releases now with no sign of immediate harm. So even if these changes don't help us resolve the issue, it might at least cut down on the number of false positives we're seeing during release testing.

This PR also ends up simplifying some of the riak_test join code a little bit.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -467,38 +467,38 @@ staged_join(Node, PNode) ->
     ?assertEqual(ok, join_with_retry(Fun)),
     ok.
 
-plan_and_commit(Node) ->
+plan_and_commit(Node, AllNodes) ->
     timer:sleep(500),
     lager:info("planning cluster join"),
     case rpc:call(Node, riak_core_claimant, plan, []) of
         {error, ring_not_ready} ->
             lager:info("plan: ring not ready"),
             timer:sleep(100),
-            plan_and_commit(Node);
+            plan_and_commit(Node, AllNodes);
         {ok, _, _} ->
             lager:info("plan: done"),
-            do_commit(Node)
+            do_commit(Node, AllNodes)
     end.
 
-do_commit(Node) ->
+do_commit(Node, AllNodes) ->
     lager:info("planning cluster commit"),
     case rpc:call(Node, riak_core_claimant, commit, []) of
         {error, plan_changed} ->
             lager:info("commit: plan changed"),
             timer:sleep(100),
             maybe_wait_for_changes(Node),
-            plan_and_commit(Node);
+            plan_and_commit(Node, AllNodes);
         {error, ring_not_ready} ->
             lager:info("commit: ring not ready"),
             timer:sleep(100),
             maybe_wait_for_changes(Node),
-            do_commit(Node);
+            do_commit(Node, AllNodes);
         {error, nothing_planned} ->
             %% Assume plan actually committed somehow
             lager:info("commit: nothing planned"),
             ok;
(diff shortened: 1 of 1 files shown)


javajolt:
@nickelization This looks very interesting. I'll test it out later tonight.

src/rt.erl
```diff
@@ -1213,9 +1209,12 @@ product(Node) ->
        true -> unknown
     end.
 
-try_nodes_ready([Node1 | _Nodes], 0, _SleepMs) ->
-    lager:info("Nodes not ready after initial plan/commit, retrying"),
-    plan_and_commit(Node1);
```
javajolt: Seems weird that we were planning and committing inside of the check for ready nodes

src/rt.erl
```diff
@@ -1187,12 +1186,9 @@ join_cluster(Nodes) ->
             %% ok do a staged join and then commit it, this eliminates the
             %% large amount of redundant handoff done in a sequential join
             [staged_join(Node, Node1) || Node <- OtherNodes],
-            plan_and_commit(Node1),
-            try_nodes_ready(Nodes, 3, 500)
+            ?assertEqual(ok, wait_until(fun() -> ok == plan_and_commit(Node1, Nodes) end))
     end,
 
-    ?assertEqual(ok, wait_until_nodes_ready(Nodes)),
```
nickelization: So `wait_until_nodes_ready` is very similar to `try_nodes_ready`. Since we're now guaranteed to call `try_nodes_ready` after every iteration of `plan_and_commit`, I believe the additional call to `wait_until_nodes_ready` is redundant.

src/rt.erl
nickelization: Yeah, definitely agree. I figured even if no other benefit ultimately comes from this PR, I at least wanted to kill off that particular bit of weirdness :)

javajolt:
Let's give it a shot to see if it helps!

JeetKunDoug:
I get the feeling that just retrying the commit part isn't going to work well - it looks like, when this happens, it just gets stuck in a loop where it continues doing the plan/commit cycle forever:

```
12:48:04.078 [info] <0.353.0> planning cluster commit
12:48:04.078 [info] <0.353.0> commit: nothing planned...why???
12:48:06.582 [info] <0.353.0> planning cluster join
12:48:06.583 [info] <0.353.0> plan: done
12:48:06.583 [info] <0.353.0> planning cluster commit
12:48:06.584 [info] <0.353.0> commit: nothing planned...why???
12:48:09.087 [info] <0.353.0> planning cluster join
```

I _think_ the cause of the "nothing planned" issue is that we are seeing a ring overwrite between repl and some other thing writing to the ring, and the `plan` part of the `plan and commit` doesn't actually work. We may have more luck, if we get to the "nothing planned" case, to actually _check_ the cluster memebership to see if our plan has taken effect and, if not, start back at the join stage and retry the whole shebang.

PR #1154, opened 2016-09-15
Fix for replication_upgrade and replication2_upgrade tests

fadushin:
Changed call from rtdev:upgrade to rt:upgrade, which has the expected signature


JeetKunDoug:
+1 9ace9ff

PR #1155, opened 2016-09-15
Used Nick's trick to use the previous riak.conf to start

fadushin:
(simulating a downgrade from a previously upgraded riak.conf).


nickelization:
+1 e1e718f

PR #1156, opened 2016-09-19
Revert changes to plan/commit retry logic

nickelization:
This is a partial revert of #1153. The changes that were made to the retry logic did not work around the problem, and instead would cause the test to fall into an infinite loop. The bug we're trying to work around seems to be occurring when the ring is overwritten by some other operation, causing the planned "join" operations to be lost. Thus, trying to plan and commit again just causes us to once again get back "nothing planned".

The changes to the `try_node_ready` retry logic should still be helpful though, so we're not reverting the entirety of #1153, just the last several commits.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -467,38 +467,38 @@ staged_join(Node, PNode) ->
     ?assertEqual(ok, join_with_retry(Fun)),
     ok.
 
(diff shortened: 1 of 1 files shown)


fadushin:
+1 f8003d6

PR #1159, opened 2016-09-20
Fix intermittent failure in w1c part of overload test (and do some misc code cleanups)

nickelization:
This PR (probably) addresses this test failure:

http://giddyup.basho.com/#/projects/riak_ee/scorecards/284/284-1884-overload-centos-6-64-memory/266291/artifacts/11106706

My hypothesis to explain the test failure is that it was caused by the way the `get_victim` function is implemented and used. I haven't been able to confirm this 100% for sure, but by tweaking the code to force picking the wrong node in `get_victim`, I was able to trigger a test failure that appeared identical.

There's no reason I can find to rule out any particular vnode when choosing the victim. The way this function is used in this test, it seems we're trying to make sure we pick a vnode on a different node from the one we're connecting to, but I removed this check and forced it to pick a vnode on the same node, and the test still passed fine.

So the place this causes trouble then is the w1c tests, because there we use n=1; if we skip the first entry in the preflist because it's on the excluded node, then we will pick a victim vnode that is not actually being read from by the get fsms, and the test will fail.

This PR also includes a couple other commits with some misc code cleanups that I added while I was working on the test.

### DIFF
--- a/tests/overload.erl
+++ b/tests/overload.erl
@@ -30,13 +30,13 @@
 -define(LIST_KEYS_RETRIES, 1000).
 -define(GET_RETRIES, 1000).
 -define(BUCKET, <<"test">>).
--define(KEY, <<"hotkey">>).
+-define(VALUE, <<"overload_test_value">>).
 -define(NORMAL_TYPE, <<"normal_type">>).
 -define(CONSISTENT_TYPE, <<"consistent_type">>).
 -define(WRITE_ONCE_TYPE, <<"write_once_type">>).
--define(NORMAL_BKV, {{?NORMAL_TYPE, ?BUCKET}, ?KEY, <<"test">>}).
--define(CONSISTENT_BKV, {{?CONSISTENT_TYPE, ?BUCKET}, ?KEY, <<"test">>}).
--define(WRITE_ONCE_BKV, {{?WRITE_ONCE_TYPE, ?BUCKET}, ?KEY, <<"test">>}).
+-define(NORMAL_BUCKET, {?NORMAL_TYPE, ?BUCKET}).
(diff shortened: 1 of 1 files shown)


fadushin:
I'd suggest maybe rebasing to one commit, just to keep the git log/history under control.

nickelization:
@fadushin I agree with squashing commits together in the case of WIP commits, or in situations where one commit is a fix to another commit in the same PR.

In this case though, each commit actually contains a separate, unrelated change, each of which stands on its own. I think if we squashed this PR into a single commit we'd actually be making the history murkier and less useful.

javajolt:
Test passes

tests/overload.erl
```diff
@@ -565,13 +571,27 @@ get_num_running_gen_fsm(Node) ->
     FsmList = [ proplists:lookup(riak_kv_get_fsm, Call) || Call <- InitCalls ],
     length(proplists:lookup_all(riak_kv_get_fsm, FsmList)).
 
-remote_vnode_queue(Idx) ->
+remote_vnode_gets_in_queue(Idx) ->
```
javajolt: We are assuming that these are only `get` messages we want to count in the queue?

tests/overload.erl
```diff
@@ -366,7 +378,7 @@ read_until_success(Node) ->
     read_until_success(C, 0).
 
 read_until_success(C, Count) ->
-    case C:get(?BUCKET, ?KEY) of
+    case C:get(<<"dummy">>, <<"dummy">>) of
```
javajolt: Clarifies that this is just any read

javajolt:
:+1: 3ba8623

tests/overload.erl
nickelization: @javajolt correct.

PR #1160, opened 2016-09-22
Fix race condition in partition_repair

nickelization:
If we just wait for the old vnode to die, we are not guaranteed that the new one will have yet been started and registered with the vnode manager, so it's possible we will end up trying to do a call into the old dead vnode in the subsequent test code. We saw a couple of test failures in giddyup recently which I believe were caused by this race condition.

To fix, we can wait for the vnode manager to return a new pid instead of just waiting for the old pid to die.


javajolt:
:+1: bff9ddc

PR #1161, opened 2016-09-25
Skip missing debug log files instead of exiting riak_test

javajolt:
`riak_test` needs to successfully complete, otherwise it will hang.  So only upload the files if they are successfully generated.  Log a message as to the results of the debug creation task.


jvoegele:
+1 6491544

PR #1162, opened 2016-09-27
Clean up get_node_debug_logs/0 by building a simple list

javajolt:
The existing code does work, but it seems cleaner to simply just add those elements to the list which is needed instead of tagging and removing values.

### DIFF
--- a/src/rtdev.erl
+++ b/src/rtdev.erl
@@ -802,17 +802,10 @@ get_node_logs() ->
 
 get_node_debug_logs() ->
     NodeMap = rt_config:get(rt_nodes),
-    Logs = rt:pmap(fun(Node) ->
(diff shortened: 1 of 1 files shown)


jvoegele:
This does feel more "Erlangy", but there is a slight difference in behavior. The previous version used `pmap` to process the logs in parallel whereas this one does them serially. Not sure if that really matters in this case, but just wanted to point that out.

javajolt:
@jvoegele I noted that, too.  Since this only runs at the end of the test, once per node under GiddyUp, I don't think it would be too bad to generate them serially.  The `riak-debug` command seems to be pretty quick.

jvoegele:
+1 460501c

PR #1163, opened 2016-09-28
Add new verify_no_datatype_siblings test

nickelization:
This currently fails because it targets a bug that has not been fixed yet. I will add a comment to this PR once we have a fix available.

This test was written in response to a specific Riak bug found in September 2016 as a result of some intermittent test failures from `test_hll`. Doing a `get` on a datatype should always cause any siblings to be merged prior to returning an object, but a scenario was found that broke that invariant.

Normally datatype values should never have dots in their metadata, but the function that removes dots before writing such values was only called when merging two different objects. The very first write of a new datatype would thus include a dot, as there would be no existing object with which to perform a merge. Additionally, the merge functions in `riak_object` depend on the absence of dots to distinguish datatypes from other values, so if siblings were present and one of the siblings was a freshly written datatype, we would not merge the two values and instead return siblings on a subsequent `get`.

NB This is not an issue when using the `fetch_type` client functions, which are normally used to retrieve datatypes; those functions cause a merge to be performed regardless of the presence of dots. To reproduce this bug we must use a plain `get` operation.

I arbitrarily chose HLLs for this test just because it's the newest datatype and hasn't had a lot of testing yet, but this could just as well be implemented with any of the CRDTs we have available to us.

### DIFF
--- a/tests/verify_no_datatype_siblings.erl
+++ b/tests/verify_no_datatype_siblings.erl
@@ -1,0 +1,102 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2016 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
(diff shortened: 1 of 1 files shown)


russelldb:
Looks go to me

+1

russelldb:
Oh, great write up, and I wish _all_ riak tests had such thorough documentation. Thank you.

PR #1164, opened 2016-10-07
Add ts_simple_select_is_null to GiddyUp [JIRA: RTS-1468]

paegun:
- implemented Riak TS SQL "WHERE IS [NOT] NULL" test.

### DIFF
--- a/tests/ts_simple_select_is_null.erl
+++ b/tests/ts_simple_select_is_null.erl
@@ -80,7 +80,7 @@ query_base(Family, Series, N) ->
     "SELECT * FROM GeoCheckin WHERE"
     " myfamily='" ++ binary_to_list(Family) ++ "'"
     " AND myseries='" ++ binary_to_list(Series) ++ "'"
-    " AND time >= 1000 AND time <= " ++ integer_to_list(N * 1000).
+    " AND time >= 1000 AND time <= " ++ integer_to_list(N * 1000 + ?SPANNING_STEP).
 
 query_all(ClusterConn, DDL, Data, Family, Series, N) ->
     Qry = query_base(Family, Series, N),
@@ -94,7 +94,14 @@ query_is_null(ClusterConn, DDL, Data, Family, Series, N, Field) ->
           " AND " ++ Field ++ " IS NULL",
     {ok, {_Fields, Rows}} = ts_util:ts_query(ClusterConn, ?TEST_TYPE, DDL, Data, Qry),
     RowsN = length(Rows),
-    ?assertNotEqual(0, RowsN),
+    %% the number of NULL rows can be determined by any non-key field being NULL
+    RowsNull = lists:foldr(fun (El, Acc) ->
+                case element(4, El) of
+                    [] -> Acc + 1;
+                    _ -> Acc
+                end
+        end, 0, Data),
+    ?assertEqual(RowsNull, RowsN),
     RowsN.
 
 query_is_not_null(ClusterConn, DDL, Data, Family, Series, N, Field) ->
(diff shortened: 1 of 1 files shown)


tests/ts_simple_select_is_null.erl
```diff
@@ -0,0 +1,134 @@
...
+    RowsN = length(Rows),
+    ?assertNotEqual(0, RowsN),
+    RowsN.
+
+query_is_null(ClusterConn, DDL, Data, Family, Series, N, Field) ->
+    Qry = query_base(Family, Series, N) ++
+          " AND " ++ Field ++ " IS NULL",
+    {ok, {_Fields, Rows}} = ts_util:ts_query(ClusterConn, ?TEST_TYPE, DDL, Data, Qry),
+    RowsN = length(Rows),
+    ?assertNotEqual(0, RowsN),
```
hmmr: You should check for exact number of records with NULL/not NULL.

tests/ts_simple_select_is_null.erl
```diff
@@ -0,0 +1,134 @@
...
+        " myint       sint64              ,"
+        " myfloat     double              ,"
+        " mybool      boolean             ,"
+        " mytimestamp timestamp           ,"
+        " PRIMARY KEY ((myfamily, myseries, quantum(time, 15, 'm')),"
+        " myfamily, myseries, time))",
+    Family = <<"family1">>,
+    Series = <<"seriesX">>,
+    N = 10,
+    Data = make_data(N, Family, Series, []),
```
hmmr, line 56: In this test, it is critical to show that empty binaries do not get conflated with NULLs. Can you add some records with `<<"">>` values in `myvarchar` field?

tests/ts_simple_select_is_null.erl
paegun: There is a comment at the head of the test module that explains why the need to check for exact set sizes for each subset is overkill, but I'm making the change regardless. It is better if we all can understand the code.

tests/ts_simple_select_is_null.erl
paegun: Great call. I'm changing `make_data` to use <<"">> instead of a fuller binary to test this edge since as you said the empty string is the critical edge to test.

hmmr: One thing that still keeps the pedant in me uneasy is the fact that the number of NULL-containing rows in this setup is equal to the number of non-NULL-containing ones. Consequently, an equality assert on the number of Expected vs Got records, will still hold even when the NULLs are conflated with `<<"">>`. The minimal change to distinguish the 5 for `count(myvarchar) where myvarchar is null` from another, visually indistinguishable, 5 for `count(myvarchar) where myvarchar is not null`, would be to change `N` to an odd number. Otherwise, I'm happy all round.

hmmr:
1d3649e +1

Basho-JIRA:
Add these tests to giddyup:
- ts_simple_query_buffers_oom
- ts_simple_select_is_null
- ts_simple_query_buffers_SUITE 

**_[posted via JIRA by Brett Hazen]**_

Basho-JIRA:
Could you please included the version(s) of riak, min and max version(s), and platform(s) [~brett] 

**_[posted via JIRA by Korrigan Clark]**_

Basho-JIRA:
Time Series isn't using the min/max versions so you can leave those blank.  TS also only supports CentOS 6.  You can see lots of previous examples in the `sql` directory. 

**_[posted via JIRA by Brett Hazen]**_

Basho-JIRA:
https://github.com/basho/giddyup/pull/204 

**_[posted via JIRA by Korrigan Clark]**_

PR #1165, opened 2016-10-11
Hinted handoff tests for Time Series.

andytill:
Test cases for a basic table where the partition key and local key are the same, and one for a table where the local key has additional columns.
1. Setup a cluster
2. Stop one node
3. Write data to a node that is still up.
4. Bring the first node back up.
5. Make a query, no gaps in the results.


macintux:
We can build on this. Good start, thanks.

+1

PR #1166, opened 2016-10-17
REL-4: Update rebar and allow deps to be locked

hazen:
Currently it is not possible to lock all of the `riak_test` dependencies... until now


korry8911:
I cloned this branch, ran:

```
make lock
rm -rf deps
make locked-deps
```

versions SHAs in rebar.config.lock and in the output of make locked-deps match up and I doubled checked one repo, mochiweb, using the CHANGES.md file to locate the date of the SHAs and match it with the code that was cloned. +1

PR #1171, opened 2016-10-24
Fixed ts_simple_select_iso8601 to include both pass and failing cases

korry8911:
So here is what I got with testing the test:

```
14:41:52.751 [info] <0.332.0> 3 - Now run the query SELECT * FROM GeoCheckin WHERE time > '2016-08-02 10:19' and time < '2016-08-02 10:20' AND myfamily = 'family1' AND myseries ='seriesX' 
14:41:52.752 [info] <0.332.0> Result is {error,{1001,<<"The upper and lower boundaries are equal or adjacent. No results are possible.">>}}
14:41:52.752 [notice] <0.2.0> ts_simple_select_iso8601 Test Run Complete pass
```

@javajolt @macintux

### DIFF
--- a/tests/ts_simple_select_iso8601.erl
+++ b/tests/ts_simple_select_iso8601.erl
@@ -31,7 +31,7 @@
 -define(LOWER, "2016-08-02 10:15:00").
 -define(UPPER, "2016-08-02 10:45:00").
 
--define(TESTS, [
+-define(PASS_TESTS, [
                 %% Test format, illustrated:
 
                 %% We expect 9 seconds between 10:19:50 and 10:20:00,
@@ -69,15 +69,17 @@
                 },
 
                 {0,
-                 {">", "2016-08-02 10:19"},
-                 {"<", "2016-08-02 10:20"}
-                },
-
-                {0,
                  {">", "2016-08-02 10:19.99"},
                  {"<", "2016-08-02 10:20"}
                 }
                ]).
+
+-define(FAIL_TESTS, [
+                {0,
+                  {">", "2016-08-02 10:19"},
+                  {"<", "2016-08-02 10:20"}
+                }
+              ]).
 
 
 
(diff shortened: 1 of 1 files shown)


tests/ts_simple_select_iso8601.erl
```diff
@@ -69,16 +69,18 @@
...
-                },
-
-                {0,
                  {">", "2016-08-02 10:19.99"},
                  {"<", "2016-08-02 10:20"}
                 }
                ]).
 
+-define(FAIL_TESTS, [
+                {0,
```
macintux: Overall the change looks good @korry8911. Please drop the tally here, however, since it'll just be clutter for these tests. (You'll have to drop the code that matches and ignores the tally too.)

korry8911:
alrighty, tally has been removed @macintux

macintux:
+1

tests/ts_simple_select_iso8601.erl
```diff
@@ -107,5 +109,17 @@ confirm() ->
...
+
+    lists:foreach(
+      fun({{Op1, String1}, {Op2, String2}}) ->
+              Qry = lists:flatten(
+                      io_lib:format(QryFmt, [Op1, String1,
+                                             Op2, String2])),
+
+              RetMsg = ts_util:single_query(Conn, Qry),
+              %% Assert that RetMsg returns a tuple with error in first place {error, {}}
+              ?assertMatch({error, {_ErrCode, _ErrMsg}}, RetMsg)
```
hazen: How about specifying the expected error code and message? Or at least the error code? You could match on it like the `Tally` in the success scenarios

hazen (changes requested):

korry8911:
@javajolt i have updated it to match on error code and message

macintux:
I suggested not being specific with the error message so we could add more error conditions without changing the code, but I'm fairly easy either way.

tests/ts_simple_select_iso8601.erl
```diff
@@ -69,16 +69,19 @@
...
-
-                {0,
                  {">", "2016-08-02 10:19.99"},
                  {"<", "2016-08-02 10:20"}
                 }
                ]).
 
+-define(FAIL_TESTS, [
+                {
+                  {1001,<<"The upper and lower boundaries are equal or adjacent. No results are possible.">>},
```
hazen: Agree with @macintux here.  We should probably just match on the error number, not the text, since the text can change. Sorry for the confusion.

korry8911:
@javajolt @macintux ok so now it only matches on the error code

hazen:
:+1:

PR #1173, opened 2016-10-27
fixed riak_control to handl riak_ts version

korry8911:
@javajolt test results:

Test Results:
## riak_control-eleveldb: pass

0 Tests Failed
1 Tests Passed


korry8911:
hmm. not sure why the git diff looks the way it does @javajolt

PR #1174, opened 2016-10-27
fixed riak_control test to handle riak_ts

korry8911:
@javajolt ok the diff looks better now, fixed riak_control to handle riak_ts version the same way it handles riak-2. and riak_ee-2.
Test results:
Test Results:
## riak_control-eleveldb: pass

0 Tests Failed
1 Tests Passed

### DIFF
--- a/tests/riak_control.erl
+++ b/tests/riak_control.erl
@@ -114,6 +114,7 @@ verify_resource(Node0, Resource) ->
     Node = rt:http_url(Node0),
     Output = os:cmd(io_lib:format("curl -s -S ~s~p", [Node, Resource])),
     lager:info("Verifying node ~p resource ~p.", [Node, Resource]),
+    lager:info("~p~n", [mochijson2:decode(Output)]),
     mochijson2:decode(Output).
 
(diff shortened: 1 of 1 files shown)


tests/riak_control.erl
```diff
@@ -114,6 +114,7 @@ verify_resource(Node0, Resource) ->
     Node = rt:http_url(Node0),
     Output = os:cmd(io_lib:format("curl -s -S ~s~p", [Node, Resource])),
     lager:info("Verifying node ~p resource ~p.", [Node, Resource]),
+    lager:info("~p~n", [mochijson2:decode(Output)]),
```
hazen: Not sure we want to leave in this debug output.  If we do, let's label it like in the above line.

hazen (changes requested):
Functionality looks good, but just has the extra output we might not want.

korry8911:
@javajolt forgot that was in there, it has been removed

hazen (approved):

hazen:
:+1:

PR #1178, opened 2016-10-31
Increased rt_retry_delay for replication_upgrade test

korry8911:
@javajolt @macintux 
I added some code to set the rt_rety_delay to 5 seconds before the test is run and then reset rt_rety_delay to the previous value.

### DIFF
--- a/tests/replication_upgrade.erl
+++ b/tests/replication_upgrade.erl
@@ -4,6 +4,13 @@
 -include_lib("eunit/include/eunit.hrl").
 
 confirm() ->
+
+    OldDelay = rt_config:get(rt_retry_delay, 1000),
+    lager:info("OldDelay is ~p", [OldDelay]),
+    rt_config:set(rt_retry_delay, 5000),
+    NewDelay = rt_config:get(rt_retry_delay),
+    lager:info("NewDelay is ~p", [NewDelay]),
+
     TestMetaData = riak_test_runner:metadata(),
     FromVersion = proplists:get_value(upgrade_version, TestMetaData, previous),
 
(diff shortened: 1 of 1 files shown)


tests/replication_upgrade.erl
```diff
@@ -4,6 +4,13 @@
 -include_lib("eunit/include/eunit.hrl").
 
 confirm() ->
+
+    OldDelay = rt_config:get(rt_retry_delay, 1000),
+    lager:info("OldDelay is ~p", [OldDelay]),
+    rt_config:set(rt_retry_delay, 5000),
```
hazen: I'm a bit concerned about hard-coding a timeout.  How about something a bit more flexible, but works just the same: `OldDelay * 5`

tests/replication_upgrade.erl
```diff
@@ -4,6 +4,13 @@
 -include_lib("eunit/include/eunit.hrl").
 
 confirm() ->
+
+    OldDelay = rt_config:get(rt_retry_delay, 1000),
+    lager:info("OldDelay is ~p", [OldDelay]),
```
hazen: How about something a bit more explicit, like "Original rt_retry_delay is..."

tests/replication_upgrade.erl
```diff
@@ -4,6 +4,13 @@
...
+    rt_config:set(rt_retry_delay, 5000),
+    NewDelay = rt_config:get(rt_retry_delay),
+    lager:info("NewDelay is ~p", [NewDelay]),
```
hazen: Matching comment for rt_retry_delay

tests/replication_upgrade.erl
```diff
@@ -93,4 +100,6 @@ confirm() ->
                                rt:log_to_nodes(Nodes, "Replication with upgraded node: ~p", [Node]),
                                replication:replication(ANodes, BNodes, true)
                        end, NodeUpgrades),
+    lager:info("Resetting Delay to is ~p", [OldDelay]),
```
hazen: Again rt_retry_delay instead of "Delay"

hazen (changes requested):
Just a few tweaks to comments and values

hazen:
FYI - There are multiple delays so we just wanted to specify which one

korry8911:
should look better now @javajolt

hazen:
:+1:

hazen (approved):

PR #1184, opened 2016-11-03
Add support for gsets

russelldb:
Customer bet365 have added gset support to riak. See https://github.com/basho/riak_kv/pull/1518 (for example.)

This PR extends the existing datatype tests to include the new GSET integration. The majority of testing of the data type itself is via quickcheck and is in the riak_dt repo.


russelldb:
Ping @binarytemple-bet365

PR #1186, opened 2016-11-04
Pass in tests to riak_test script as comma separated values

korry8911:
Added some code to riak_test_escript to allow passing in of comma separated test values after the -t flag. Example usage would be:
```
 ./riak_test -c rtdev -t test_a,test_b,test_c
```

If you pass in repeated tests, the test is only run once. The following examples will run test_a once:
```
 ./riak_test -c rtdev -t test_a,test_a,test_a
 ./riak_test -c rtdev -t test_a,test_b,test_a
 ./riak_test -c rtdev -t test_b,test_a,test_a
```

tests must be comma separated with no whitespace in between.
@javajolt

### DIFF
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -116,7 +116,6 @@ main(Args) ->
                                           {lager_file_backend, [{file, "log/test.log"},
                                                                 {level, ConsoleLagerLevel}]}]),
     lager:start(),
-
     %% Report
     Report = case proplists:get_value(report, ParsedArgs, undefined) of
         undefined -> undefined;
@@ -135,8 +134,7 @@ main(Args) ->
     CommandLineTests = parse_command_line_tests(ParsedArgs),
     Tests0 = which_tests_to_run(Report, CommandLineTests),
     lager:info("Running Tests: ~p", [[TestName || {_, {TestName,_}} <- Tests0]]),
-
-  case Tests0 of
+    case Tests0 of
         [] ->
             lager:warning("No tests are scheduled to run"),
             init:stop(1);
@@ -220,8 +218,6 @@ parse_command_line_tests(ParsedArgs) ->
     [code:add_patha(CodePath) || CodePath <- CodePaths,
                                  CodePath /= "."],
     Dirs = proplists:get_all_values(dir, ParsedArgs),
-    lager:info("Dirs: ~p", [Dirs]),
-
     SkipTests = string:tokens(proplists:get_value(skip, ParsedArgs, []), [$,]),
     DirTests = lists:append([load_tests_in_dir(Dir, SkipTests) || Dir <- Dirs]),
     lists:foldl(fun(Test, Tests) ->
@@ -240,13 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
         end, [], lists:usort(DirTests ++ SpecificTests)).
 
 extract_test_names(Test, {CodePaths, TestNames}) ->
+
+    RTTestPaths = get_test_paths(),
     CommaSepTests = string:tokens(Test, [$,]),
-    WCTests = [TestA || TestA <- CommaSepTests, string:str(TestA, "*") > 0],
-    NotWCTest = [TestB || TestB <- CommaSepTests, string:str(TestB, "*") =:= 0],
-    AllTests = lists:append([filelib:wildcard(TestC) || TestC <- WCTests])++NotWCTest,
+
+    WCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= yes_wc_no_prefix],
+    WCTestsPrefix = [TestA || TestA <- CommaSepTests, get_test_type(TestA) =:= yes_wc_yes_prefix],
+    NotWCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_no_prefix],
+    NotWCTestsPrefix = [TestA || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_yes_prefix],
+
+    AllWCTests = lists:append([filelib:wildcard(TestC) || TestC <- WCTestsNoPrefix++WCTestsPrefix]),
+    AllNotWCTests = NotWCTestsNoPrefix++NotWCTestsPrefix,
+    AllTests = AllWCTests++AllNotWCTests,
+
     CodePathList = [filename:dirname(TestName) || TestName <- AllTests],
     TestNameList = [filename:rootname(filename:basename(TestName)) || TestName <- AllTests],
     {CodePathList++CodePaths, TestNameList++TestNames}.
+
+get_test_type(Test) ->
+    case string:str(Test, "*") of
+        0 ->
+          case string:str(Test, "/") of
+              0 -> no_wc_no_prefix;
+              _ ->  no_wc_yes_prefix
+          end;
+        _ ->
+          case string:str(Test, "/") of
+              0 -> yes_wc_no_prefix;
+              _ -> yes_wc_yes_prefix
+          end
+    end.
+
+get_test_paths() ->
+    RTTestPaths = rt_config:get(test_paths, ["tests"]),
+    case RTTestPaths of
+        ["tests"] ->
+            RTTestPaths;
+        _ -> RTTestPaths++["tests"]
+    end.
+
+get_test_with_valid_prefix(Test, []) ->
+    Test;
+
+get_test_with_valid_prefix(Test, TestPaths) ->
+    [FirstPath|Rest] = TestPaths,
+    case filelib:wildcard(FirstPath++"/"++Test) of
+          [] -> get_test_with_valid_prefix(Test, Rest);
+          _  -> FirstPath++"/"++Test
+    end.
 
 which_tests_to_run(undefined, CommandLineTests) ->
     {Tests, NonTests} =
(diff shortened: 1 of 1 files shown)


andytill:
@korry8911 great that this is being looked at! Requiring that these are comma separated means that it is not possible to use bash file wild cards. For example, if I want to run all the tests for Riak Time Series I would want to run `-t tests/ts_*` and have the tests be all of the arguments after `-t`.

hazen:
@korry8911 So Andy brings up a good point.  It would be nice to support wildcards, too.  Beyond that, it would be nice to have the parser allow spaces between the test names as a convenience, e.g.
`-t ts_simple_select, ts_cluster_comprehensive`

korry8911:
@andytill so it seems that with wildcards, ts_\* would be expanded into a whitespace separated list.  The library used in riak_test_escript is getopt and has a restriction on white spaces: https://github.com/jcomellas/getopt#arguments-with-embedded-whitespace. To use white spaces, the argument must be enclosed by \", like \"test_a test_b test_c\". I'm pretty sure using bash wildcards does not work within quotes. But, there is probably a way to expand the wildcards from within erlang.  Let me know if you have any ideas or suggestions.

korry8911:
@javajolt @andytill so a temp work around could be to use:

```
./riak_test -c rtdev -t $(echo tests/ts_* | sed 's/ /,/g')
./riak_test -c rtdev -t $(echo tests/ts_cluster* | sed 's/ /,/g')
```

but be careful since ts_\* wil pick up files that are not tests like ts_ops, so being more specific with the wildcards is the way to go

andytill:
getopt supports non-option args at the end  of the command line which could be used for test suites, however this is already being used for `HarnessArgs`.

https://github.com/basho/riak_test/blob/be0f135b7d56d93bf486e85bc46ac2f994f0975f/src/riak_test_escript.erl#L68

I don't think this is actually used anywhere so could be re-purposed, but you'd have to double check.

This bash snippet can be used to run multiple tests currently in rt.

``` shell
./riak_test -c rtdev `sh -c 'for i in tests/ts_*; do echo -t $i; done'`
```

korry8911:
@andytill @javajolt i think the problem with using HarnessArgs would be that the tests being passed in will be split among two different variables:

```
./riak_test -c rtdev -t testA testB testC
```

testA will be in the ParsedArgs and then testB and testC will be in harness args.... I don't see a clean way to reconcile the two, but there is a way. If this is the prefered option i can modify the code to assume arguments in harness args should be added to the test args. In fact, we could just drop the -t flag all together and make harness args the default way to pass in tests. Thought?

andytill:
I like it, I am in favour of repurposing `HarnessArgs` as long as we don't break anything. Maybe keep the `-t` option around for a while until people get used to a life without it :)

korry8911:
@andytill @javajolt ok so i reworked this to support passing in comma separated lists which can contain wildcards. There two caveats. 1) if you pass in a single wildcard, you must enclose in quotes, this is due to shell completing the expression, and 2) if you pass in wildcards, you must pass in the test with its directory. The following are examples running from /riak_test:

```
BAD USAGE
./riak_test -c rtdev -t ts_simple_* %% no tests run
./riak_test -c rtdev -t "ts_simple_* %% no tests run
./riak_test -c rtdev -t ts_simple_*,yz_handoff %% only yz_handoff run

GOOD USAGE
./riak_test -c rtdev -t "tests/ts_simple_*" %% will run all matching tests from ./tests directory
./riak_test -c rtdev -t tests/ts_simple_*,yz_handoff %% will run all matching tests in ./test and yz_handoff
./riak_test -c rtdev -t "tests/ts_simple_*","tests/ts_cluster_*",yz_handoff 
./riak_test -c rtdev -t tests/ts_simple_*,"tests/ts_cluster_*",yz_handoff 
./riak_test -c rtdev -t tests/ts_simple_*,tests/ts_cluster_*,yz_handoff
```

korry8911:
@andytill @javajolt ok, so I made it much better. You can now pass in really anything. No quotes needed for anything. The script will now search for tests using wildcards in the directories specified in riak_config for test_paths. If it cant find the tests in there is makes one last try in the `./tests/` directory in riak_test. If you supply a prefix it will just use that. So now here are example usages:

```
GOOD USAGE
./riak_test -c rtdev -t ts_cluster_*
./riak_test -c rtdev -t tests/ts_cluster_*
./riak_test -c rtdev -t ts_cluster_*,ts_simple_*,yz_handoff
```

hazen:
How exciting! I can't wait to try it out

src/riak_test_escript.erl
```diff
@@ -236,8 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
         end, [], lists:usort(DirTests ++ SpecificTests)).
 
 extract_test_names(Test, {CodePaths, TestNames}) ->
-    {[filename:dirname(Test) | CodePaths],
-     [filename:rootname(filename:basename(Test)) | TestNames]}.
+
+    RTTestPaths = get_test_paths(),
+    CommaSepTests = string:tokens(Test, [$,]),
+
+    WCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= yes_wc_no_prefix],
```
hazen: Looks like by building 4 lists you are touching each test 4 times.  How about move through the list once, but for each element in the perform different logic.  Maybe first check to see if it has a wildcard.  If so, then expand the list of test names.  You could use `lists:foldl` to build an accumulator of additional tests.  Before adding all of those tests to the list, then you could compare each new one to see if it has a prefix and see if it needs to be added to the code path and stripped from the test name.

Happy to discuss to clarify what I'm suggesting.

src/riak_test_escript.erl
```diff
@@ -236,8 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
...
+          end;
+        _ ->
+          case string:str(Test, "/") of
+              0 -> yes_wc_no_prefix;
+              _ -> yes_wc_yes_prefix
+          end
+    end.
+
+get_test_paths() ->
+    RTTestPaths = rt_config:get(test_paths, ["tests"]),
```
hazen: How about we simply have a default value of `[]` and return the value of `rt_config:get/2` concatenated with `tests`?

`[rt_config:get(test_paths, [])|["tests"]]`

src/riak_test_escript.erl
```diff
@@ -236,8 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
...
+    case RTTestPaths of
+        ["tests"] ->
+            RTTestPaths;
+        _ -> RTTestPaths++["tests"]
+    end.
+
+get_test_with_valid_prefix(Test, []) ->
+    Test;
+
+get_test_with_valid_prefix(Test, TestPaths) ->
```
hazen: You could use `lists:foldl` instead of do-it-yourself recursion.

src/riak_test_escript.erl
```diff
@@ -236,8 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
...
+    NotWCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_no_prefix],
+    NotWCTestsPrefix = [TestA || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_yes_prefix],
+
+    AllWCTests = lists:append([filelib:wildcard(TestC) || TestC <- WCTestsNoPrefix++WCTestsPrefix]),
+    AllNotWCTests = NotWCTestsNoPrefix++NotWCTestsPrefix,
+    AllTests = AllWCTests++AllNotWCTests,
+
+    CodePathList = [filename:dirname(TestName) || TestName <- AllTests],
+    TestNameList = [filename:rootname(filename:basename(TestName)) || TestName <- AllTests],
+    {CodePathList++CodePaths, TestNameList++TestNames}.
```
hazen, line 242: List concatenation is more efficient than the `++` operator, as in the original code

src/riak_test_escript.erl
```diff
@@ -236,8 +236,54 @@ parse_command_line_tests(ParsedArgs) ->
...
+    WCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= yes_wc_no_prefix],
+    WCTestsPrefix = [TestA || TestA <- CommaSepTests, get_test_type(TestA) =:= yes_wc_yes_prefix],
+    NotWCTestsNoPrefix = [get_test_with_valid_prefix(TestA, RTTestPaths) || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_no_prefix],
+    NotWCTestsPrefix = [TestA || TestA <- CommaSepTests, get_test_type(TestA) =:= no_wc_yes_prefix],
+
+    AllWCTests = lists:append([filelib:wildcard(TestC) || TestC <- WCTestsNoPrefix++WCTestsPrefix]),
+    AllNotWCTests = NotWCTestsNoPrefix++NotWCTestsPrefix,
+    AllTests = AllWCTests++AllNotWCTests,
+
+    CodePathList = [filename:dirname(TestName) || TestName <- AllTests],
```
hazen: We can probably combine this so it only passes over the tests once.  Actually we might be able to build a list of tuples of code path and test and `lists:unzip/1` them when it's time to return the values.

hazen (changes requested):
Code works great! However there are some optimizations we can add to make it a bit more efficient and compact.

korry8911:
@javajolt ok I added some of the suggestions you made... the only exception I think is that I use ++ still in some locations because we are concatenating two lists and not just adding an element to the head of a list.

korry8911:
@javajolt @andytill ok, so it should be good to go know, barring any more comments...

src/riak_test_escript.erl
```diff
@@ -258,11 +258,11 @@ process_tests(Test, Acc) ->
 get_test_with_valid_prefix(Test, []) ->
     Test;
 
-get_test_with_valid_prefix(Test, TestPaths) ->
```
hazen: I like it!

hazen (approved):

hazen:
:+1:

andytill:
Nice 👍

PR #1187, opened 2016-11-04
Break ts_util into three utility modules

hazen:
Rest assured that this is only the first step to cleaning up some tests.  @macintux's work makes what test are doing clearer, especially around cluster construction and test setup.

That said, @andytill has a good point about moving some of the data into the tests, so next round we should look at that as well.


korry8911:
+1

andytill:
Discussion against this change in https://github.com/basho/riak_test/pull/1187

PR #1188, opened 2016-11-10
Prevent tree upgrades which can block replication from completing

JeetKunDoug:
Use capability override to make sure that neither side
of the replication link upgrades its object hash version
because if once side updates its trees and the other hasn't
completed the process yet (or in mixed clusters doesn't even
support it) the AAE full sync cannot complete.


bsparrow435:
Manually tested locally and confirmed in logs that capability stays as `legacy`. Merging.

PR #1189, opened 2016-11-10
Use rt:wait_until to check unacked on source node

ian-mi:
Acks are sent asynchronously after a successful write on the sink, so we
cannot rely on the ack having being received by the source immediately
after we read the data on the sink.


JeetKunDoug:
+1 - looks good, makes sense, test passes on local testing.

PR #1190, opened 2016-11-11
Add rt:wait_until to final counter check

bsparrow435:
Added a rt:wait_until for the final check of the counters instead of
only checking once. Sometimes, even though the full sync completed
successfully, the counter values might not be fully written and
propagated to all owning vnodes. Change test to check more than once to
account for this eventual consistency.


JeetKunDoug:
+1 - makes sense, passes locally.

PR #1192, opened 2016-11-11
Removed superfluous call repair after being cancelled. [JIRA: RIAK-2970]

fadushin:
The verify_2i_aae test fails in some cases due to an apparent race condition between restarting a cancelled repair and grabbing the Pid of the `riak_kv_2i_aae` `gen_fsm`.

This PR removes this added repair after cancellation, as the test is not checking anything that the requested repairs actually occurred.


JeetKunDoug:
+1

bashopatricia:
create jira issue

PR #1201, opened 2016-11-17
DESCRIBE <table> changes for 1.5

paegun:
* fixed ts_simple_describe_table test to match change in the column header "Is Null" => "Nullable", "Primary Key" => "Partition Key" and addition of "Sort Order" column.

Related code change:
* https://github.com/basho/riak_ql/commit/be1ac4ad3be3072ec9eb8dbea5bcaa4a452ed3ae#diff-f16ee8e3e139a9b6c0c9663962632df4R34


hazen:
:+1: 3128eda

PR #1208, opened 2016-11-23
[RTS-1094] Tests for check table readiness on decode

paegun:
* added test for case that SELECT is executed for a table that does not exist.

Depends on:
* https://github.com/basho/riak_kv/pull/1545


gordonguthrie:
+1 77eb57c

PR #1209, opened 2016-11-23
Query buffer suite chmod fix kc

korry8911:
@javajolt @hmmr 
ok so this test needs to remove and restore write permission to the query buffer directory to test that the proper error is passed. The way it was doing this is through chmod but since riak_test is run using root user in basho_builds, chmod is rendered useless. To regain the expect behavior there are two function chattr and chflags that are now used.  The reason for using two different functions is that linux uses one and osx uses the other.  As far as I can tell these are the only two functions that will remove and restore write permissions for root user for their respectifve OSes. I inlcuded the option of using the OSX chlags so that local testing will have the same behavior as the basho_builds runs.

### DIFF
--- a/tests/ts_simple_query_buffers_SUITE.erl
+++ b/tests/ts_simple_query_buffers_SUITE.erl
@@ -389,11 +389,11 @@ col_no({[A|_], _, _}) ->
 
 modify_dir_access(take_away, QBufDir) ->
     ct:pal("take away access to ~s\n", [QBufDir]),
-    file:rename(QBufDir, QBufDir++".boo"),
-    file:write_file(QBufDir, "nothing here");
+    ok = file:rename(QBufDir, QBufDir++".boo"),
+    ok = file:write_file(QBufDir, "nothing here");
 
 modify_dir_access(give_back, QBufDir) ->
     ct:pal("restore access to ~s\n", [QBufDir]),
-    file:delete(QBufDir),
-    file:rename(QBufDir++".boo", QBufDir).
+    ok = file:delete(QBufDir),
+    ok = file:rename(QBufDir++".boo", QBufDir).
 
--- a/tests/ts_simple_query_buffers_SUITE.erl
+++ b/tests/ts_simple_query_buffers_SUITE.erl
@@ -216,10 +216,9 @@ init_per_testcase(query_orderby_max_data_size_error, Cfg) ->
 init_per_testcase(query_orderby_ldb_io_error, Cfg) ->
     Node = hd(proplists:get_value(cluster, Cfg)),
     QBufDir = filename:join([rtdev:node_path(Node), "data/query_buffers"]),
-    OSType = os:cmd("uname"),
-    Cmd = get_write_perm_cmd(take_away, OSType) ++  fmt(" ~s", [QBufDir]),
-    CmdOut = "" = os:cmd(Cmd),
-    ct:log("~s: '~s'", [Cmd, CmdOut]),
+    Cmd = get_write_perm_cmd(take_away, QBufDir),
+    ct:log("running ~s", [Cmd]),
+    "" = os:cmd(Cmd),
     Cfg;
 
 init_per_testcase(_, Cfg) ->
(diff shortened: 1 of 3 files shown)


korry8911:
this pr also includes changes to ts_simple_aggregation_fail and ts_simple_aggregation.  Previously timestamps were not allowed in agg functions but now that is fine. I moved a timestamp agg function test that used to fail to the passing tests since that is now acceptable behaviour.

korry8911:
furthermore, I still need to test the query buffers test on linux... works fine on osx though

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -238,9 +237,9 @@ end_per_testcase(query_orderby_max_data_size_error, Cfg) ->
 end_per_testcase(query_orderby_ldb_io_error, Cfg) ->
     Node = hd(proplists:get_value(cluster, Cfg)),
     QBufDir = filename:join([rtdev:node_path(Node), "data/query_buffers"]),
-    Cmd = fmt("chmod +w '~s'", [QBufDir]),
-    CmdOut = "" = os:cmd(Cmd),
-    ct:log("~s: '~s'", [Cmd, CmdOut]),
```
hazen, line 243: Capturing the command output seems useful. Can we leave that in?

korry8911: sure, will put it back in

hazen:
@hmmr I like the solution of moving files rather than OS-specific commands

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -393,3 +392,10 @@ col_no({[A|_], _, _}) ->
 
 fmt(F, A) ->
     lists:flatten(io_lib:format(F, A)).
+
+get_write_perm_cmd(take_away, QBufDir) ->
+    ct:pal("take away perms on ~s\n", [QBufDir]),
+    fmt("sh -c \"mv '~s' '~s.boo' && touch '~s'\"", [QBufDir, QBufDir, QBufDir]);
```
hazen: Rather than using the shell, what about native Erlang like `file:rename/2` and `file:write/2`?  Or back to the original permission solution with `file:change_mode/2`?

hmmr: I have no strong preference here. It's just a very expressive one-liner bash job compared with double (gasp) the number of lines when done using native Erlang means. It would matter if we run the tests on a Windows box (which we are not going to do I suppose :)

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -218,7 +218,8 @@ init_per_testcase(query_orderby_ldb_io_error, Cfg) ->
     QBufDir = filename:join([rtdev:node_path(Node), "data/query_buffers"]),
     Cmd = get_write_perm_cmd(take_away, QBufDir),
     ct:log("running ~s", [Cmd]),
-    "" = os:cmd(Cmd),
+    CmdOut = "" = os:cmd(Cmd),
```
hmmr: No, you want to print it before matching to `""`.

hmmr (changes requested):

tests/ts_simple_query_buffers_SUITE.erl
hazen: Good point

korry8911:
@hmmr @javajolt switched the shell code to erlang, test should be good to go

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -392,12 +386,14 @@ col_no({[A|_], _}) ->
...
 
-get_write_perm_cmd(take_away, QBufDir) ->
-    ct:pal("take away perms on ~s\n", [QBufDir]),
-    fmt("sh -c \"mv '~s' '~s.boo' && touch '~s'\"", [QBufDir, QBufDir, QBufDir]);
-get_write_perm_cmd(give_back, QBufDir) ->
-    ct:pal("restore perms on ~s\n", [QBufDir]),
-    fmt("sh -c \"rm '~s' && mv '~s.boo' '~s'\"", [QBufDir, QBufDir, QBufDir]).
+modify_dir_access(take_away, QBufDir) ->
+    ct:pal("take away access to ~s\n", [QBufDir]),
+    file:rename(QBufDir, QBufDir++".boo"),
```
hazen: Do we want to match on results to crash quickly, e.g., if the directory is missing or permissions are messed up, etc? `ok = file:rename(...`

korry8911:
@javajolt  added some ok's

hmmr (approved):

hazen:
```
Test Results:
ts_simple_query_buffers_SUITE-riak_kv_bitcask_backend,riak_kv_eleveldb_backend: pass
ts_simple_aggregation-riak_kv_bitcask_backend,riak_kv_eleveldb_backend        : pass
ts_simple_aggregation_fail-riak_kv_bitcask_backend,riak_kv_eleveldb_backend   : pass
0 Tests Failed
3 Tests Passed
```
:+1:

PR #1217, opened 2016-11-30
merge/fix for Added regression tests for sql delete (feature-gg-delete)

paegun:
* merged riak_ts-develop.
* fixed empty results now return a textual response.
* fixed result output is (has been) csv, not tabular.

TODO: DELETE should also return a textual response like INSERT, ie "Deleted 1 row.", see https://github.com/basho/riak_shell/pull/62/files#diff-8d41632d1b8366d9b85c50744267130eR105


hazen:
Wonder if we should just have @gordonguthrie rebase his original PR?

paegun:
These changes were addressed in the upstream branch, so closing.

PR #1218, opened 2016-12-01
RTS-1466: Add ts_simple_object_size_limits

hazen:
Requires https://github.com/basho/riak_kv/pull/1505

Basically the limits for object sizes have been reduced for TS and extended to be imposed on the write-once path.


macintux:
+1

I think it'd be useful to define medium-sized objects and check for the expected log messages, but that's far from critical for now.

hazen:
Scraping the log for the warnings does seem like a nice-to-have, but is a bit ugly.  Let's do this for starters.  We can expand the test once it's in.

PR #1219, opened 2016-12-03
Fix for ts_simple_recompile_ddl regression

paegun:
* fixed ts_simple_recompile_ddl, using riak_kv_compile_tab encapsulation to reduce the chance and cost of regression break.

depends on: https://github.com/basho/riak_kv/pull/1558

### DIFF
--- a/tests/ts_simple_recompile_ddl.erl
+++ b/tests/ts_simple_recompile_ddl.erl
@@ -19,14 +19,13 @@
 %% -------------------------------------------------------------------
 
 -module(ts_simple_recompile_ddl).
-
 -behavior(riak_test).
 
 -include_lib("eunit/include/eunit.hrl").
 
 -export([confirm/0]).
--define(LEGACY_TABLE, riak_kv_compile_tab).
--define(DETS_TABLE, riak_kv_compile_tab_v2).
+-define(PREVIOUS_TABLE, riak_kv_compile_tab_v2).
+-define(CURRENT_TABLE, riak_kv_compile_tab_v3).
 
 confirm() ->
     [Node | _Rest] = Cluster = ts_setup:start_cluster(1),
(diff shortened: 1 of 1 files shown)


tests/ts_simple_recompile_ddl.erl
```diff
@@ -36,49 +35,58 @@ confirm() ->
...
-    %% 3) A table which seemingly was stuck in the compiling state
-    ok = dets:insert(?LEGACY_TABLE, {<<"Table1">>, Table1DDL, Pid, compiled}),
-    ok = dets:insert(?DETS_TABLE, {<<"Table2">>, 1, Table2DDL, Pid2, compiled}),
-    ok = dets:insert(?DETS_TABLE, {<<"Table3">>, 1, Table3DDL, Pid3, compiling}),
-    dets:close(?LEGACY_TABLE),
-    dets:close(?DETS_TABLE).
+    %% 1) An old DETS entry in compiled state
+    %% 2) An old DETS entry in compiling state
+    %% 3) A current, valid DETS entry
+    ok = riak_kv_compile_tab:insert_previous(<<"Table1">>, Table1DDL, compiled),
```
javajolt: I'm wondering if we actually want to call KV directly here or mimic the logic.  Just thinking about interfaces to the code.

hazen: It looks like the pattern currently used for running code on the Riak nodes is
```
PL = rpc:call(Node, riak_core_apl, get_primary_apl, [DocIdx, NVal, riak_kv]),
```
Since I'd like riak_test to run on remote nodes, this might be safer than calling the KV version on the riak_test node.

paegun:
@javajolt much appreciated on the call to `rpc:call` and encapsulate more.

hazen:
:+1:

PR #1221, opened 2016-12-05
ts_updown_util: filter timeseries.max_running_fsms in convert_riak_conf

hmmr:
With a newly added parameter `timeseries.max_running_fsms` to riak_kv.schema, it needs to be filtered when converting riak.conf for 1.4 nodes in the updowngrade scenarios.


gordonguthrie:
+1

PR #1224, opened 2016-12-06
Give table activation a better chance to succeed

macintux:
We'll periodically see a test failure when `ts_setup:activate_bucket_type` gives up on table activation before the cluster is ready. Decrease the odds of this occurring by gradually increasing the delay between activation attempts.


andytill:
Shouldn't those tests call `rt:wait_until_ring_converged/1` to make sure the cluster is ready before performing operations on it and expecting those operations to retry?

macintux:
It happened today on a single node cluster, so it's not purely a question of ring convergence.

andytill:
+1

PR #1226, opened 2016-12-06
Update the replication parameters which are now TS defaults

hazen:
Changes to replication values changed the expected outputs https://bashoeng.atlassian.net/browse/RTS-1469 and https://github.com/basho/riak_core/pull/870


macintux:
+1

PR #1227, opened 2016-12-07
always use current client in updowngrade tests

hmmr:
There are no changes in wire protocol between 1.4 and 1.5, meaning both 1.4 and 1.5 clients should continue to work with 1.5 as well as mixed clusters. This PR turns off `use_previous_client` flag to always use the current (1.5) client during up/downgrade scenarios.


hazen:
FINALLY got this test to run...and finish

hazen:
:+1:

PR #1228, opened 2016-12-08
use ts stats when testing ts node

korry8911:
Added the option to use TS specific stats when testing a TS node and KV specific stats when testing a KV node. @javajolt


hazen:
+1

PR #1230, opened 2016-12-08
test for NULLs in single-key put in ts_simple_http_api

hmmr:
Test for NULLs in single-key put in ts_simple_http_api. Also, ensure `rt:wait_for_service` has a chance to succeed.


paegun:
+1

hazen:
+1

PR #1232, opened 2016-12-08
Add HTTP null put test, SELECT ... IS NULL and SHOW TABLES tests.

paegun:
* thanks to number_or_null, added HTTP null value put, null value in select, and show tables tests.

for #1230

### DIFF
--- a/tests/ts_simple_http_api_SUITE.erl
+++ b/tests/ts_simple_http_api_SUITE.erl
@@ -62,10 +62,12 @@ all() ->
     [ create_table_test,
       create_bad_table_test,
       create_existing_table_test,
+      show_tables_test,
       describe_table_test,
       describe_nonexisting_table_test,
       bad_describe_query_test,
       post_single_row_test,
+      post_single_row_with_null_test,
       post_single_row_missing_field_test,
       post_single_row_wrong_field_test,
       post_several_rows_test,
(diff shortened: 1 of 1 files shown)


tests/ts_simple_http_api_SUITE.erl
```diff
@@ -211,6 +226,13 @@ select_test(Cfg) ->
         "\"rows\":[[\"q1\",\"w1\",11,110],"
         "[\"q1\",\"w1\",20,119]]}" = Body.
 
+select_with_null_test(Cfg) ->
+    Select = "select * from bob where a='qN' and b='wN' and c>1 and c<99 and d is null",
+    {ok,"200", Headers, Body} = execute_query(Select, Cfg),
+    "application/json" = content_type(Headers),
+    "{\"columns\":[\"a\",\"b\",\"c\",\"d\"],"
+        "\"rows\":[[\"qN\",\"wN\",11,[]]]}" = Body.
```
hmmr, line 234: Shouldn't that be `null` instead of `[]`? Such as here: https://github.com/basho/riak_test/pull/1230/files#diff-fab6193d0172e564f9e34f4cda49fefdR358.

paegun: `null` is for the JSON, ?SQL_NULL `[]` is for RiakTS (erlang).

hmmr:
+1

tests/ts_simple_http_api_SUITE.erl
hmmr: You are correct, again.

PR #1235, opened 2016-12-09
The existing failure logging for some reason is not printing out

gordonguthrie:
the failures to the ct results page or the logs

Additional logging has been added in a belt-and-braces style


andytill:
+1

PR #1236, opened 2016-12-10
[RTS-1298] Fix ts_simple_http_security_SUITE HTTP put of NULL.

paegun:
* fixed ts_simple_http_security_SUITE to pass `null` over HTTP, was using `[]` which is an empty array in JSON.

Locally the test now passes, but that is strange given stated lack of support for HTTP security yet in Riak TS. Please verify that the test does not pass. If it does, *surprise* we have HTTP security.

### DIFF
--- a/tests/ts_simple_http_security_SUITE.erl
+++ b/tests/ts_simple_http_security_SUITE.erl
@@ -175,7 +175,7 @@ make_data() ->
     [{list_to_binary(fmt("A~5..0b", [I rem 2])),
       <<"B">>,
       I + 1,
-      if I rem 5 == 0 -> []; el/=se -> I end}  %% sprinkle some NULLs
+      if I rem 5 == 0 -> null; el/=se -> I end}  %% sprinkle some NULLs
      || I <- lists:seq(1, 300)].
 make_data_ann(Columns, Data) ->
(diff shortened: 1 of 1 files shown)


javajolt:
I'm still getting some unrelated local failure:
```
=== reason = {assertMatch_failed,
                     [{module,ts_simple_http_security_SUITE},
                      {line,141},
                      {expression,"rhc_ts : put ( Client , ? TABLE1 , Data )"},
                      {pattern,"ok"},
                      {value,
                          {error,
                              {400,
                               <<"Bad value for field \"d\" of type sint64 in table \"t1\"">>}}}]}
  in function  ts_simple_http_security_SUITE:'-put_test/1-fun-1-'/2 (tests/ts_simple_http_security_SUITE.erl, line 141)
  in call from ts_simple_http_security_SUITE:put_test/1 (tests/ts_simple_http_security_SUITE.erl, line 141)
  in call from test_server:ts_tc/3 (test_server.erl, line 1358)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 975)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 924)
```

hazen:
:+1:

PR #1237, opened 2016-12-12
Wait for Webmachine to be up before running tests

andytill:
There are lots of failures like the following on the nightly test.

screen shot 2016-12-12 at 10 48 50

The error is that WebMachine is not ready. Later tests have 404s because they rely on tables created by other tests!  Webmachine is not a "service" like `riak_kv` or `riak_pipe` so I have coded this like as a poll making a request.


gordonguthrie:
+1

hazen:
This seems like it might be useful for other tests using webmachine.  Perhaps we should put it in a utility module? Also we should see what other tests are currently doing about this same issue.

PR #1239, opened 2016-12-12
Make the replay out put handle unicode input

gordonguthrie:


gordonguthrie:
Goes with this `riak-shell` PR

https://github.com/basho/riak_shell/pull/68

andytill:
Can you describe what this change is, I don't fully understand?

hazen:
@gordonguthrie `tts_cluster_riak_shell_regression_log` seems to fail with this change so we either need to fix the test to accommodate this change, or skip this change

PR #1242, opened 2016-12-12
Fix flappy pb_security test

paegun:
* fix flappy pb_security test by retrying on pb connection timeout, proposing such a function be factored down into rt.


hmmr:
+1

hazen:
+1

PR #1244, opened 2016-12-12
Fix flappy repl_cancel_fullsync

paegun:
* fix flappy repl_cancel_fullsync, making it less likely for the replication to complete before the cancel request is received.

Run w/o this change 10 executions of the test were found to be consistently yielding a fail. Run w/ this change 20 executions of the test were found to succeed. AND the test execution is not significantly increased.

### DIFF
--- a/tests/repl_cancel_fullsync.erl
+++ b/tests/repl_cancel_fullsync.erl
@@ -5,7 +5,8 @@
 
 -define(TEST_BUCKET,
         <<"repl-cancel-fullsync-failures-systest_a">>).
--define(NUM_KEYS, 1000).
+%% increased from 1000 to 10000 b/c replication is completing before cancel sometimes (flappy test)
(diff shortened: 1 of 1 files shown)


hmmr:
+1

With a note to posterity: should it start failing again due to fullsync completing before the test attempts to interrupt it, bump the limit again. (Or perhaps use intercepts and dilate time at certain points down there.)

paegun:
@hmmr The comments were for the purpose of suggesting such a thing, flappy test:
* `increased from 1000 to 10000 b/c replication is completing before cancel sometimes (flappy test)`
* `%%<< intentionally half of ?NUM_KEYS`

Yes, I agree an intercept would likely ensure the flappiness would not occur.

hazen:
Let's give it a shot!

hazen:
+1

PR #1245, opened 2016-12-13
Test new base64 handling for HTTP requests

macintux:
Convert one of the `varchar` fields in the standard table to a `blob`.

Requires https://github.com/basho/riak_kv/pull/1572


gordonguthrie:
+1 - we should probably use this underlying test as a refactoring example...

hazen:
:+1:

PR #1247, opened 2016-12-13
Check stats on all nodes for REPL fullsync success

andytill:
The `server_fullsyncs` stat only gets incremented on the node that executed the fullsync. Previously, the test however checked one of the nodes in cluster which makes it sometimes pass, sometimes fail.

This change means all the nodes in the list of nodes which should make up the cluster are checked, if any of the nodes has the incremented value then the fullsync is successful.


paegun:
+1

PR #1248, opened 2016-12-14
Fix ts_cluster_riak_shell_regression_log race on activation, slow build boxes

paegun:
*DO NOT MERGE*

* fixed ts_cluster_riak_shell_regression_log race condition on activation exposed by slow build boxes.

Please see HACK comment in ts_cluster_riak_shell_regression_log.erl


gordonguthrie:
The CREATE TABLE only returns success if the table is activated

https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_svc.erl#L269

So the flappyness is caused by a different bug somewhere in the body of `riak_core`

paegun:
@gordonguthrie good point, so CREATE TABLE is determining that the table is active, but the INSERT query is failing to determine the same. Splitting the log in 2 hacks around this, but better to chase down the gap between the 2 methods of determining if the table is "Active" (ready for query).

The gap follows:
* riak_kv_ts_util:is_table_active_and_supported/2, which additionally ensures:
  * the DDL module is compiled
  * the capability is at min version supported on the cluster

I propose we close this hack and change the CREATE TABLE successful exit condition to be based on riak_kv_ts_util:is_table_active_and_supported/2, not simply bucket-type activation.

PR #1249, opened 2016-12-14
Add Gradle tests for S3 API

jbrisbin:
Add a test that delegates to the Gradle-based tests in the `s3-api-tests` repo (a copy of which is expected to be checked out next to `riak_test`). It runs the gradle tests and checks the output for any lines that end in `FAILED`, which would indicate a test failure. If any are found, the test is considered to have failed. Output from the Gradle test run is logged to info level. Still need to add some logic to start riak with the S3 API using the profile backend for testing.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -279,13 +279,16 @@ rpc_get_env(Node, [{App,Var}|Others]) ->
 
 -spec connection_info(node() | [node()]) -> interfaces() | conn_info().
 connection_info(Node) when is_atom(Node) ->
-    {ok, [{PB_IP, PB_Port}]} = get_pb_conn_info(Node),
-    {ok, [{HTTP_IP, HTTP_Port}]} = get_http_conn_info(Node),
-    case get_https_conn_info(Node) of
-        undefined ->
-            [{http, {HTTP_IP, HTTP_Port}}, {pb, {PB_IP, PB_Port}}];
-        {ok, [{HTTPS_IP, HTTPS_Port}]} ->
-            [{http, {HTTP_IP, HTTP_Port}}, {https, {HTTPS_IP, HTTPS_Port}}, {pb, {PB_IP, PB_Port}}]
+    {ok, [PB_Info]} = get_pb_conn_info(Node),
+    {ok, [HTTP_Info]} = get_http_conn_info(Node),
+    Info0 = [{pb, PB_Info}, {http, HTTP_Info}],
+    Info1 = case get_https_conn_info(Node) of
+        undefined -> Info0;
+        {ok, [{HTTPS_IP, HTTPS_Port}]} -> [{https, {HTTPS_IP, HTTPS_Port}} | Info0]
+    end,
+    case get_s3_conn_info(Node) of
+        undefined -> Info1;
+        {ok, [{S3_IP, S3_Port}]} -> [{s3, {S3_IP, S3_Port}} | Info1]
     end;
 connection_info(Nodes) when is_list(Nodes) ->
     [ {Node, connection_info(Node)} || Node <- Nodes].
@@ -319,6 +322,15 @@ get_http_conn_info(Node) ->
 get_https_conn_info(Node) ->
     case rpc_get_env(Node, [{riak_api, https},
                             {riak_core, https}]) of
+        {ok, [{IP, Port}|_]} ->
+            {ok, [{IP, Port}]};
+        _ ->
+            undefined
+    end.
+
+-spec get_s3_conn_info(node()) -> [{inet:ip_address(), pos_integer()}].
+get_s3_conn_info(Node) ->
+    case rpc_get_env(Node, [{riak_s3_api, http}]) of
         {ok, [{IP, Port}|_]} ->
             {ok, [{IP, Port}]};
         _ ->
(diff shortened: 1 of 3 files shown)


src/rt.erl
```diff
@@ -279,13 +279,16 @@ rpc_get_env(Node, [{App,Var}|Others]) ->
...
-        undefined ->
-            [{http, {HTTP_IP, HTTP_Port}}, {pb, {PB_IP, PB_Port}}];
-        {ok, [{HTTPS_IP, HTTPS_Port}]} ->
-            [{http, {HTTP_IP, HTTP_Port}}, {https, {HTTPS_IP, HTTPS_Port}}, {pb, {PB_IP, PB_Port}}]
+    {ok, [PB_Info]} = get_pb_conn_info(Node),
+    {ok, [HTTP_Info]} = get_http_conn_info(Node),
+    Info0 = [{pb, PB_Info}, {http, HTTP_Info}],
+    Info1 = case get_https_conn_info(Node) of
+        undefined -> Info0;
+        {ok, [{HTTPS_IP, HTTPS_Port}]} -> [{https, {HTTPS_IP, HTTPS_Port}} | Info0]
```
jbrisbin, line 287: I still think this could be less noisy by collapsing the tuple.

src/rt.erl
```diff
@@ -279,13 +279,16 @@ rpc_get_env(Node, [{App,Var}|Others]) ->
...
+    end,
+    case get_s3_conn_info(Node) of
+        undefined -> Info1;
+        {ok, [{S3_IP, S3_Port}]} -> [{s3, {S3_IP, S3_Port}} | Info1]
```
jbrisbin, line 291: Same here ^^^

paulhenrich:
Nice. I was able to run locally, see the failure, and see the output of the gradle tests. 👍

jbrisbin:
Depends on basho/s3-api-tests#6

tests/s3_api_gradle.erl
```diff
@@ -1,33 +1,54 @@
...
-    %% Set to testing profile backend on all the nodes
-    {[ok, ok, ok], []} = rpc:multicall(Nodes,
-                                       application,
-                                       set_env,
-                                       [riak_s3_api, default_user_backend, riak_s3_user_profile_backend]),
+    %% Create test user and wait for it to appear on the cluster
+    _ = rpc:call(Node, riak_s3_user, create, ?TEST_USER),
+    rt:wait_until(fun() ->
+                        {Results, []} = rpc:multicall(Nodes, riak_core_metadata, get, ?METADATA_KEY),
+                        case lists:filter(fun(R) -> R == undefined end, Results) of
```
JeetKunDoug: Unless I'm misunderstanding something, you really want to do something like:
```erlang
lists:all(fun(R) -> R =/= undefined end, Results)
```
instead of the case statement here... you want to wait until all nodes have the metadata key you're looking for, and none of them return undefined, right?  As it stands now, it'll return `false` if there are no records that are undefined (which means keep waiting) but `true` otherwise (when there are still undefined results)?

tests/s3_api_gradle.erl
```diff
@@ -1,33 +1,54 @@
...
+                            [] -> false;
+                            _ -> true
+                        end
+                  end),
+    [
+        {"s3.access_key_id", S3AccessKey},
+        {"s3.display_name", _},
+        {"s3.secret_access_key", S3SecretKey}
+    ] = rpc:call(Node, riak_core_metadata, get, ?METADATA_KEY),
```
JeetKunDoug, line 34: Are we guaranteed to get the 3 parameters (s3.access_key_id/display_name/secret_assecc_key) in the same order every time, on every platform? Otherwise, I'd sort the result and the list your pattern-matching on just to be sure.

JeetKunDoug (commented):
2 questions, but otherwise looks nice... should work well.

tests/s3_api_gradle.erl
```diff
@@ -0,0 +1,68 @@
...
+    confirm/0
+]).
+
+-define(NODE_COUNT, 3).
+-define(TEST_USER, ["test", "test@example.com"]).
+-define(METADATA_KEY, [{<<"security">>, <<"users">>}, <<"test">>]).
+
+confirm() ->
+    %% Build a simple cluster
+    Nodes = rt:build_cluster(?NODE_COUNT),
```
JeetKunDoug, line 20: We'll need to force the use of the `riak_kv_elevedb_backend` in this test before long, because only that backend will (initially at least) support the necessary `grouped_list_keys` functionality we'll need to do a good list keys test.

This can be accomplished by passing a Config parameter to build_cluster that includes the backend specifically - something like this:

```erlang
 Config = [{riak_kv, [
        {storage_backend, Backend}]}
    ],
  Nodes = rt:build_cluster(?NODE_COUNT, Config),
...
```

hazen:
@jbrisbin What are all of the prerequisites to run this test? Just thinking about how we plan to automate it.  We'd definitely need to add the `s3-api-tests` repository and just gradle?

jbrisbin:
@javajolt It only needs Java and a clone of the basho/s3-api-tests repo alongside the `riak_test` dir (though we could automate that as well, I suppose). Gradle is downloaded as-needed by the wrapper (we're using `./gradle` rather than just `gradle`).

paulhenrich:
👍

hazen:
We'll have to rethink this so that OSS users can still build RT. We are looking at conditionally compiling riak EE, so we might want to take the same approach.  This is fine for a branch, but will need love if we want to merge to develop.

paulhenrich:
thumbot +1

paulhenrich:
thumbot merge

PR #1251, opened 2016-12-14
redbug tracing in riak test

andytill:
This change adds the module `rt_redbug` which is a thin wrapper around redbug, a trace can be set like the following.

```erlang
rt_redbug:trace('dev1@127.0.0.1', "riak_ql_parser:parse")
```

By default, this trace will not be set.  By adding `--trace` to the `riak_test` command arguments, traces in the test suite will be set.

### DIFF
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -214,12 +214,8 @@ parse_command_line_tests(ParsedArgs) ->
     %% this is a little ugly, the process that creates the ets table to store
     %% the tracing state cannot shutdown or the table will be gc'ed, so create
     %% a dummy proc that doens't die
-    proc_lib:spawn(
-        fun() ->
-            rt_redbug:new(),
-            rt_redbug:set_tracing_applied(proplists:is_defined(apply_traces, ParsedArgs)),
-            timer:sleep(infinity)
-        end),
+    rt_redbug:new(),
+    rt_redbug:set_tracing_applied(proplists:is_defined(apply_traces, ParsedArgs)),
     %% Parse Command Line Tests
     {CodePaths, SpecificTests} =
         lists:foldl(fun extract_test_names/2,
--- a/src/riak_test_escript.erl
+++ b/src/riak_test_escript.erl
@@ -44,7 +44,8 @@ cli_options() ->
  {upgrade_version,    $u, "upgrade",  atom,       "which version to upgrade from [ previous | legacy ]"},
  {keep,        undefined, "keep",     boolean,    "do not teardown cluster"},
  {report,             $r, "report",   string,     "you're reporting an official test run, provide platform info (e.g. ubuntu-1204-64)\nUse 'config' if you want to pull from ~/.riak_test.config"},
- {file,               $F, "file",     string,     "use the specified file instead of ~/.riak_test.config"}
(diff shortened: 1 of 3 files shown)


hazen:
@andytill Did you get the trace arguments working?

andytill:
@javajolt yes.

hmmr:
+1

src/rt_redbug.erl
```diff
@@ -0,0 +1,110 @@
...
+-export([set_tracing_applied/1]).
+-export([stop/1]).
+-export([trace/2, trace/3]).
+
+%% Create a new ets to store globally whether we're tracing or not.
+new() -> 
+    rt_trace_tab = ets:new(rt_trace_tab, [named_table]).
+
+set_tracing_applied(TracingApplied) when is_boolean(TracingApplied) ->
+	lager:info("Setting apply tracing to ~p", [TracingApplied]),
```
hazen: @andytill Could you please replace your tabs with spaces? Cheers!

src/riak_test_escript.erl
```diff
@@ -210,6 +211,15 @@ parse_command_line_tests(ParsedArgs) ->
                    [] -> [undefined];
                    UpgradeList -> UpgradeList
                end,
+    %% this is a little ugly, the process that creates the ets table to store
+    %% the tracing state cannot shutdown or the table will be gc'ed, so create
+    %% a dummy proc that doens't die
```
hazen, line 216: I'd investigate using `rt_config` instead of an ets table for global config state

src/rt_redbug.erl
andytill: Fixed in 4e1da92

src/riak_test_escript.erl
andytill: Fixed in d337dfc

src/riak_test_escript.erl
```diff
@@ -44,7 +44,8 @@ cli_options() ->
  {upgrade_version,    $u, "upgrade",  atom,       "which version to upgrade from [ previous | legacy ]"},
  {keep,        undefined, "keep",     boolean,    "do not teardown cluster"},
  {report,             $r, "report",   string,     "you're reporting an official test run, provide platform info (e.g. ubuntu-1204-64)\nUse 'config' if you want to pull from ~/.riak_test.config"},
- {file,               $F, "file",     string,     "use the specified file instead of ~/.riak_test.config"}
+ {file,               $F, "file",     string,     "use the specified file instead of ~/.riak_test.config"},
+ {apply_traces,undefined, "trace",    undefined,  "Apply traces to the target node, defined in the SUITEs"}
```
gordonguthrie, line 48: Not sure what the 'Short Codes' are used for...

`apply_traces` looks like verbose semantically so it prolly should have a short code...

gordonguthrie: Duh, short codes means short codes, Brexit means Brexit and Gordon means idiot...

andytill: My thoughts behind not having a short code was that tracing was weird enough that typing `--tracing` was warranted, plus `-t` is taken.

gordonguthrie:
+1

PR #1252, opened 2016-12-15
get qbuf dir directly from riak_kv env

hmmr:
When "removing" query buffer root dir to emulate IO error in query_orderby_ldb_io_error, a default value for that dir was used. For the cases when it is set to some other value in riak.conf (or similar), we need to correctly read it from application env.


hmmr:
Incorporated in https://github.com/basho/riak_test/pull/1257.

PR #1253, opened 2016-12-16
Allow redbug to be enabled from the config file

hazen:
Now you can also set `{apply_traces, true}` in the `.riak_test.config` file

### DIFF
--- a/src/rt_redbug.erl
+++ b/src/rt_redbug.erl
@@ -31,13 +31,17 @@
 -export([trace/2, trace/3]).
 
 set_tracing_applied(TracingApplied) when is_boolean(TracingApplied) ->
-    %% If this flag is set on the command line, but not in the config file
-    %% then switch it on.
-    Enabled = case rt_config:get(apply_traces, undefined) of
-        undefined -> TracingApplied;
-        ConfigValue -> ConfigValue
+    %% Enable redbug tracing if enabled via command-line or config file
+    Enabled = case TracingApplied of
+        true -> TracingApplied;
+        _ -> rt_config:get(apply_traces, false)
     end,
-    lager:info("Setting apply tracing to ~p", [Enabled]),
+    case Enabled of
+        true ->
+            lager:warning("Will enable any redbug traces contained in the test");
+        _ ->
+            lager:warning("Will not enable any redbug traces contained in the test")
+    end,
     rt_config:set(apply_traces, Enabled).
 
 is_tracing_applied() ->
--- a/src/rt_redbug.erl
+++ b/src/rt_redbug.erl
@@ -26,21 +26,22 @@
 -module(rt_redbug).
 
 -export([is_tracing_applied/0]).
--export([new/0]).
 -export([set_tracing_applied/1]).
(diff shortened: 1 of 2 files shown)


hmmr:
+1

src/rt_redbug.erl
```diff
@@ -26,21 +26,22 @@
...
 set_tracing_applied(TracingApplied) when is_boolean(TracingApplied) ->
-    lager:info("Setting apply tracing to ~p", [TracingApplied]),
-    rt_config:set(apply_traces, TracingApplied).
+    %% If this flag is set on the command line, but not in the config file
+    %% then switch it on.
+    Enabled = case rt_config:get(apply_traces, undefined) of
+        undefined -> TracingApplied;
+        ConfigValue -> ConfigValue
+    end,
+    lager:info("Setting apply tracing to ~p", [Enabled]),
```
macintux: Would like to see this message be more explicit and higher level (`warning`?) given the potential impact on performance if tracing is enabled without the user realizing it.

Perhaps `Will enable any redbug traces contained in the test` (and `Will not enable any redbug traces contained in the test`) for the true/false messages.

src/rt_redbug.erl
```diff
@@ -26,21 +26,22 @@
...
-%% Create a new ets to store globally whether we're tracing or not.
-new() -> 
-    ok.
-
 set_tracing_applied(TracingApplied) when is_boolean(TracingApplied) ->
-    lager:info("Setting apply tracing to ~p", [TracingApplied]),
-    rt_config:set(apply_traces, TracingApplied).
+    %% If this flag is set on the command line, but not in the config file
+    %% then switch it on.
+    Enabled = case rt_config:get(apply_traces, undefined) of
```
macintux: Doesn't this allow the config (when `false`) to override the command line? I think you want to use `false` instead of `undefined` as your default value and first case, but haven't thought it entirely through.

hazen: Would you prefer favoring the command-line?  The `undefined` value simply denotes it is not set in the config file.

src/rt_redbug.erl
hazen: This was just Andy's logging comment.  Happy to change it if you'd like.

src/rt_redbug.erl
andytill: Typically command line arguments always overrule configuration.

andytill:
Why bother with default configuration values? If it is not set anywhere then the value is false as it is currently in develop.

macintux:
That's an excellent point. I can't see myself turning tracing on by default, which is the only use case here.

src/rt_redbug.erl
hazen: OK, i'll switch it. Since it's not a boolean, you can only enable it via the command-line, but not disabled it.

hazen:
The real motivation for this change is so we could enable this in a 1-off GiddyUp test run to aid in debugging.  Config file generation is much more easily automated than command-line options.

macintux:
+1

PR #1256, opened 2016-12-23
Consolidate all artifacts into a single zip file

hazen:
Also move more artifact reporting over to `giddyup.erl`

Had to switch from a tar file to a ZIP file because of the 100-character filename restriction of the Erlang implementation of tar.


korry8911:
+1

PR #1257, opened 2016-12-24
test in-memory ORDER BY

hmmr:
RTS-1641

Depends on https://github.com/basho/riak_kv/pull/1587.

New test case to check the delayed leveldb ops in query buffers, with rudimentary performance stats.

### DIFF
--- a/tests/ts_simple_query_buffers_SUITE.erl
+++ b/tests/ts_simple_query_buffers_SUITE.erl
@@ -165,7 +165,7 @@ init_per_testcase(query_orderby_max_data_size_error, Cfg) ->
 
 init_per_testcase(query_orderby_ldb_io_error, Cfg) ->
     Node = hd(proplists:get_value(cluster, Cfg)),
-    QBufDir = filename:join([rtdev:node_path(Node), "data/query_buffers"]),
+    QBufDir = get_qbuf_dir(Node),
     modify_dir_access(take_away, QBufDir),
     Cfg;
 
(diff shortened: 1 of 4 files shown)


macintux:
Testing `nulls_{first|last}` would seem to be a useful improvement.

tests/ts_simple_query_buffers_SUITE.erl
```diff
@@ -167,48 +167,129 @@ init_per_testcase(query_orderby_max_data_size_error, Cfg) ->
...
+
+    ok = Res,
+    ok.
 
 
+%%
+%% 4. checking transition from inmem to ldb during data collection
+%% --------------------------------------
+
+query_orderby_inmem2ldb(Cfg) ->
```
macintux, line 249: I may be missing something here, but it would seem useful to verify that the in-memory times are significantly less than the disk times, to see that memory is actually being used. (Not sure how to define "significant" to avoid flappy tests.) This test doesn't seem like it's actually testing anything, just displaying measurements.

macintux:
I'd also like to see a test to verify that the on-disk buffers are actually being cleaned up after the appropriate interval.

tests/ts_simple_query_buffers_SUITE.erl
hmmr: The test case in https://github.com/basho/riak_test/pull/1257/files#diff-095dac84f424d6a5f5dd43a19ac4c37eR257 emulates the conditions to cause the qbuf manager to behave functionally differently: (1) immediately write qbufs to leveldb, (2) hold the data completely in memory and (3) hold for some time, then dump it. Cases (1) and (2) are assured by the intercepted `can_afford_inmem` always returning `false` or `true`, respectively.

The naive measurements I provide are for illustrative purposes; as expected, the pure-inmem trial completes faster, but it is perf team business to estimate by how much.

hmmr:
Added a test case to check all temp tables are deleted after all qbufs expire, in 35151e8.

(Also, rebased the whole thing on integration branch.)

macintux:
+1

hmmr:
Belated answer to https://github.com/basho/riak_test/pull/1257#issuecomment-275197010: all of them are actually tested in `query_orderby_comprehensive`, as specified both implicitly and explicitly.

PR #1261, opened 2017-01-11
Merge sweeper test code to develop [JIRA: RIAK-3237]

nickelization:
This is a rebase of Mikael's obj_ttl branch onto develop, along with a few fixes I added in recent months.


go.sh
```diff
@@ -0,0 +1,7 @@
+./riak_test -c riak_ee_dev -t repl_fs_bench3 -D'fs_num_keys=1000000' -D'fs_strategy=keylist' -D'fs_direct=0'
```
hazen: Why are we checking this into `riak_test`? Seems like this is someone's local test script.

nickelization: I honestly don't know. It was on the branch Mikael used to develop the sweeper tests, so I assumed it was relevant, but maybe he somehow had some extra commits on his branch that shouldn't have been there. I'll try and see if we can get rid of these jtuple commits without disturbing the rest of the sweeper work....

nickelization: @javajolt I rebased the riak_test sweeper branch and got rid of those two commits, and everything else still seems to work. Just pushed the updated branch.

nickelization:
One note I forgot to add here: sweeper_long_test and sweeper_bench do not work out of the box. The other tests are the important ones that we've used for testing the latest sweeper code in 2.3, and those two older tests are left over from Mikael's sweeper work. I left them in the PR in case anyone wants to refer back to them someday, and maybe even fix them up and get them working again.

bashopatricia:
create jira issue

JeetKunDoug:
+1

PR #1264, opened 2017-01-20
Update build versions and remove git --local options

hazen:
Older versions of git, like those shipped with platforms we are using for testing like CentOS 6 do not support `git config --local`

```
git --version
git version 1.7.1
git config --local user.name "Riak Test"
error: unknown option `local'
```


korry8911:
+1

PR #1268, opened 2017-01-26
Test alter_table with bucket properties

macintux:
In order to make leveldb-based expiry easier to leverage for users, we are exposing bucket properties via a new `alter_table` command.

Update the `ts_simple_show_create_table` test to include alter_table.

Requires https://github.com/basho/riak_kv/pull/1616 and associated PRs

### DIFF
--- a/tests/ts_simple_show_create_table.erl
+++ b/tests/ts_simple_show_create_table.erl
@@ -28,20 +28,18 @@
 
 confirm() ->
     Table = "ShowCreateTable",
-    DDL = "CREATE TABLE " ++ Table ++ " ("
-    " somechars   VARCHAR   NOT NULL,"
-    " somebool    BOOLEAN   NOT NULL,"
-    " sometime    TIMESTAMP NOT NULL,"
-    " somefloat   DOUBLE,"
-    " PRIMARY KEY ((somechars, somebool, QUANTUM(sometime, 1, 'h')), "
-    " somechars, somebool, sometime)) "
-    " WITH (n_val=2)",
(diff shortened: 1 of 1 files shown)


paegun:
+1

tests/ts_simple_show_create_table.erl
```diff
@@ -53,13 +59,25 @@ confirm() ->
...
+
+initialize(Table) ->
+    DDL = "CREATE TABLE " ++ Table ++ " ("
+    " somechars   VARCHAR   NOT NULL,"
+    " somebool    BOOLEAN   NOT NULL,"
+    " sometime    TIMESTAMP NOT NULL,"
+    " somefloat   DOUBLE,"
+    " PRIMARY KEY ((somechars, somebool, QUANTUM(sometime, 1, 'h')), "
+    " somechars, somebool, sometime)) "
+    " WITH (n_val=2)",
```
hazen, line 79: Should this not also be a parameter, instead of hard-coded, if you changed it to one in `expected_show/2`? I guess it does not change on initialization...

hazen (commented):
Is this the right place for this test?  It was designed for `show table`, but now we've added `alter table`

macintux:
We need a way to verify the results, so a test which already does a create/show table seemed a logical place to extend to do create/show/update/show.

tests/ts_simple_show_create_table.erl
macintux: Yeah, meh. For a test I'm not worried about some redundancy in that regard. I mainly wanted to consolidate the output strings.

hazen:
+1

PR #1275, opened 2017-02-28
delete qbuf_inmem_max from riak.conf when downgrading to 1.5

hmmr:
RTS-1801

When downgrading, care should be taken to delete from riak.conf settings not supported by the previous version. In the 1.6 vs 1.5 cycle, this concerns one setting: `riak_kv.query.timeseries.qbuf_inmem_max`. This PR updates `convert_riak_conf_to_previous` for the current cycle.


hazen:
:+1:

PR #1276, opened 2017-02-28
Fix test failures caused by riak client not allowing list keys

hazen:
Mostly simply the addition of the `riak-erlang-client` environment variable to allow listing of buckets and keys (now disabled by default).  A couple of actual test changes.


hmmr:
+1

PR #1277, opened 2017-02-28
Add EQC test for list group keys

jvoegele:
This EQC test is a riak_test that verifies the list group keys functionality provided by https://github.com/basho/riak_kv/pull/1620

The primary goal of this test is to ensure that paginated results are correct and consistent even in the face of varying coverage plans for the individual page requests. As such, the test initiates each request against a random node in the test cluster, and also uses EQC generators to test against various values for the `max_keys` parameter. It also tests against several different buckets that each have different `n_val`s, in the range of 1 to 5.

### DIFF
--- a/tests/kv_list_group_keys_eqc.erl
+++ b/tests/kv_list_group_keys_eqc.erl
@@ -31,7 +31,6 @@ confirm() ->
     quickcheck(?MODULE:prop_group_keys_prefix(Cluster)),
     quickcheck(?MODULE:prop_group_keys_delimiter(Cluster)),
     quickcheck(?MODULE:prop_group_keys_prefix_delimiter(Cluster)),
-    rt:clean_cluster(Cluster),
     pass.
 
 %% ====================================================================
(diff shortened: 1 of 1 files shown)


tests/kv_list_group_keys_eqc.erl
```diff
@@ -0,0 +1,198 @@
...
+
+%% ====================================================================
+%% EQC Properties
+%% ====================================================================
+
+quickcheck(Property) ->
+    ?assert(eqc:quickcheck(eqc:numtests(500, Property))).
+
+forall_buckets_and_max_keys(Property) ->
+    ?FORALL(Bucket, oneof(buckets()),
```
nickelization, line 51: Not critiquing, but just curious, since I'm not quite an expert-level EQC user myself: what's the purpose of using the nested `?FORALL` statements instead of just doing something like:
```
?FORALL({Bucket, MaxKeys}, {oneof(buckets()), gen_max_keys()},
    collect(MaxKeys, Property(Bucket, MaxKeys))).
```

nickelization:
+1 - Doug mentioned to me about the "keys" vs "objects" change, and that makes sense to me. Other than that, test code all looks solid :)

PR #1279, opened 2017-03-01
Ensure `rt:connection_info` returns it's list in sorted order

JeetKunDoug:
Several tests currently fail because of changes in `connection_info`
which resulted in the list being returned not being in order any more.
Some tests expect `http` to be first, and some assume the result can be
used as an `orddict` directly.

### DIFF
--- a/src/rt.erl
+++ b/src/rt.erl
@@ -279,17 +279,20 @@ rpc_get_env(Node, [{App,Var}|Others]) ->
 
 -spec connection_info(node() | [node()]) -> interfaces() | conn_info().
 connection_info(Node) when is_atom(Node) ->
-    {ok, [PB_Info]} = get_pb_conn_info(Node),
-    {ok, [HTTP_Info]} = get_http_conn_info(Node),
-    Info0 = [{pb, PB_Info}, {http, HTTP_Info}],
-    Info1 = case get_https_conn_info(Node) of
-        undefined -> Info0;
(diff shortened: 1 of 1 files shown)


src/rt.erl
```diff
@@ -279,17 +279,20 @@ rpc_get_env(Node, [{App,Var}|Others]) ->
...
-    {ok, [PB_Info]} = get_pb_conn_info(Node),
-    {ok, [HTTP_Info]} = get_http_conn_info(Node),
-    Info0 = [{pb, PB_Info}, {http, HTTP_Info}],
-    Info1 = case get_https_conn_info(Node) of
-        undefined -> Info0;
-        {ok, [{HTTPS_IP, HTTPS_Port}]} -> [{https, {HTTPS_IP, HTTPS_Port}} | Info0]
+    {ok, [HTTP_Info0]} = get_http_conn_info(Node),
+    HTTP_Info = [{http, HTTP_Info0}],
+    {ok, [PB_Info0]} = get_pb_conn_info(Node),
+    PB_Info = [{pb, PB_Info0}],
```
fadushin: Not to be totally anal, but maybe list PB_Info after HTTPS_Info, since, you know, we are wanting it to be ordered (I realize we achieve order by the ++ below)

fadushin:
+1

paulhenrich:
+1

PR #1281, opened 2017-03-01
Merge latest TS 1.6.x back to develop

hazen:
This is mostly to address the breaking change of the Erlang client to not support list key and list buckets out of the box


fadushin:
+1

PR #1282, opened 2017-03-02
Merge develop to s3-develop to pick up riak client changes

JeetKunDoug:


hazen:
I read this as from `s3` to `develop` and got nervous for a minute

PR #1283, opened 2017-03-03
updowngrade: better prepare riak.conf & caps, make it work for 1.4-1.5 and 1.5-1.6 cycles

hmmr:
RTS-1661?

Make code version-agnostic, only require newly introduced keys and caps (those that we need to wait_until before actually performing the tests) to be registered when stepping to next release up/downgrade cycle. This should reduce the patch size when upgrading the updowngrade machinery for the next cycle.

```
ts_cluster_updowngrade_group_by_SUITE-eleveldb          : pass
ts_cluster_updowngrade_order_by_SUITE-eleveldb          : pass
ts_cluster_updowngrade_select_aggregation_SUITE-eleveldb: pass
```

### DIFF
--- a/tests/ts_updown_util.erl
+++ b/tests/ts_updown_util.erl
@@ -588,7 +588,7 @@ get_riak_release_in_slot(VsnSlot) ->
         unknown ->
             ct:fail("Failed to determine riak version in '~s' slot", [VsnSlot]);
         Known ->
-            case re:run(Known, "riak_ts-(\\d+)\.(\\d+)\.(\\d+)", [{capture, all_but_first, list}]) of
+            case re:run(Known, "riak_ts-(\\d+)\\.(\\d+)\\.(\\d+)", [{capture, all_but_first, list}]) of
                 {match, [V1, V2, V3]} ->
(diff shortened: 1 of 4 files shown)


tests/ts_updown_util.erl
```diff
@@ -581,31 +582,87 @@ wait_until_active_table(TargetNode, PrevClientNode, Table, N) ->
 make_msg(Format, Payload) ->
     list_to_binary(fmt(Format, Payload)).
 
+
+get_riak_release_in_slot(VsnSlot) ->
+    case rtdev:get_version(VsnSlot) of
+        unknown ->
+            ct:fail("Failed to determine riak version in '~s' slot", [VsnSlot]);
+        Known ->
+            case re:run(Known, "riak_ts-(\\d+)\.(\\d+)\.(\\d+)", [{capture, all_but_first, list}]) of
```
macintux: The `\.` should be `\\.`

macintux:
Other than the regular expression comment, this looks really good.

hmmr:
+1

paegun:
+1

PR #1285, opened 2017-03-07
Add w1c delete tests

erikleitch:
This PR is in support of the KV PR to optimize the write-once delete path by removing the unnecessary read before delete (https://github.com/basho/riak_kv/pull/1630).

This PR adds a separate test to verify all versions of write-once deletes, including:

* deletes via the `riakc_pb_socket:delete` api from write-once KV buckets
* deletes via the `riakc_ts:delete` api from TS tables
* deletes via the `riakc_ts:query` api from TS tables

### DIFF
--- a/tests/ts_verify_w1c_delete.erl
+++ b/tests/ts_verify_w1c_delete.erl
@@ -1,0 +1,189 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2015 Basho Technologies, Inc.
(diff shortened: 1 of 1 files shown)


erikleitch:
(and I've verified that the test passes with both the current riak_ts-develop branch and the eml/optimize_w1c_delete branch of riak_kv)

hazen:
:+1:

PR #1288, opened 2017-03-13
Test for leveldb bucket-level expiry in EE

macintux:
Requires https://github.com/basho/riak_core/pull/904 and https://github.com/basho/eleveldb/pull/235.


macintux:
Note to @javajolt: EE only in giddyup

macintux:
Also: in order to test this locally, you need to `export BASHO_EE=1` before building. This will make sure `leveldb` gets its EE bits.

PR #1289, opened 2017-03-13
correctly parse {1,2,3} out of "riak_ts_ee-1.2.3"

hmmr:
Better parse version string to allow it to be "riak_ts_ee-1.2.3" in addition to "riak_ts-1.2.3".


hazen:
:+1:

PR #1292, opened 2017-03-13
Make the test keyspace unique for prop_failure_replication_test [JIRA: RTS-1816]

hazen:
There could eventually be some keyspace crosstalk from the previous test `no_ts_real_time_replication_test`, which had been using the same keys.  Once realtime replication is enabled, perhaps those keys were eventually getting written to Cluster B.  Now just use an independent range, just in case.


macintux:
+1

DSomogyi:
create jira issue

PR #1293, opened 2017-03-15
Set force_hashtree_upgrade to true

bsparrow435:
This removes the race condition where some nodes are upgrading and others arent in the middle of an active fullsync. We are not testing AAE upgrades with this test so we should disable them.


JeetKunDoug:
+1

PR #1295, opened 2017-03-16
RTS-1820: Disable fullsync when clusters are disconnected

hazen:


macintux:
It's possible this will lead to test failures, but if so, the tests themselves are broken, not this change.

+1

PR #719, opened 2014-12-30
Riak Tests for scenarios of basho/riak_kv#679

russelldb:
Note that the fallback test `kv679_dataloss_fb` will fail until that case is addressed with a riak_object format change.

PR #1296, opened 2017-03-21
Re-add old kv679 test and add new test for further dataloss edge case

russelldb:
`kv679_dataloss_fb.erl` test was part of the original kv679 suite, but at the time of 2.1
it was put in it's own branch. See https://github.com/basho/riak_test/pull/719 for discussion and decision. In short, Riak Test is made up of passing tests,
and kv679_dataloss_fb.erl still failed. Adding this back to a
branch off develop as I know have a fix of this issue.

There is also a new test, `kv679_dataloss_fb2.erl`, that shows another dataloss edge case in riak, even in 2.1 with
per-key-actor-epochs enabled. The test is a little convoluted, and is
based on a quickcheck counter example, included in the riak_kv/test
directory. In short, both this test, and the other kv679_dataloss_fb
test, show that even with multiple replicas acking/storing, a single
disk error on a single replica is enough to cause acked writes to be
silently and permanently lost. For a replicated database, that is bad

### DIFF
--- a/intercepts/riak_kv_bitcask_backend_intercepts.erl
+++ b/intercepts/riak_kv_bitcask_backend_intercepts.erl
@@ -48,3 +48,11 @@ corrupting_get(Bucket, Key, ModState) ->
             
 corrupt_binary(O) ->
     crypto:rand_bytes(byte_size(O)).
+
+always_corrupt_get(Bucket, Key, ModState) ->
+    case ?M:get_orig(Bucket, Key, ModState) of
+        {ok, BinVal0, UpdModState} ->
+            BinVal = corrupt_binary(BinVal0),
+            {ok, BinVal, UpdModState};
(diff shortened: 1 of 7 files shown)


russelldb:
I'm going to be opening the PR against riak_kv that provides a fix for both these tests in a short while, and will edit to provide link when I do. In the meantime, feel free to run these against 2.1 and up.

nickelization:
Both tests make sense, and I confirmed they both fail without basho/riak_kv#1643 and pass with it. Still reviewing the riak_kv PR, but the test PR looks good to me. Thanks, Russell!

nickelization:
+1

JeetKunDoug:
+1 - merging to fix `partition_repair` before next Giddyup run. Still need to add the two new tests to the Giddyup database.

PR #1299, opened 2018-02-22
Add a test for `node_confirms` as put option

russelldb:
Test for https://github.com/basho/riak_kv/pull/1663

(NOTE: as usual with riak-test you will have to faff with your deps to get those branches that you need. We really need a better way to do this.)


martinsumner:
Already merged into `develop-2.9` and `develop-3.0`.  OK to merge to `2.2.5` here for completeness

PR #1305, opened 2018-03-29
Override default retry delay

russelldb:
The test times out if the retry delay is enabled. Also, add some
notification as to which of the many tests is running for better log
spelunking


ghost (approved):
+1

PR #1306, opened 2018-04-05
These ensemble tests passed by chance

russelldb:
In 2.2.3 and before the ring created by claim for these ensemble tests
was dangerous, but by good fortune the tests passed. The claim/ring
fixes in 2.2.5 cause the tests to fail. Upping the ring-size to 64
allows the tests to pass legitimately, and with a more realistic
scenario. Sadly they are also slower.


ghost (approved):
+1

PR #1307, opened 2018-06-06
try merging develop-2.2.5 to develop ? 

ghost:


russelldb:
I have a lot of things in nhs-riak/develop-2.2.5 still to PR over to basho/develop-2.2.5. After which, we should create a 2.2.5 tag of riak-test, for posterity/regression testing of 2.2.5 series.

Be careful about riak-test develop, since we don't currently use (or plan to use) riak develop.

Just my pennyworth…

ghost:
ok - just putting it out there for feedback - thanks.

PR #1314, opened 2019-03-18
Develop 2.9

martinsumner:


tests/http_security.erl
```diff
@@ -505,10 +510,9 @@ confirm() ->
                                          {verify, verify_peer},
                                          {reuse_sessions, false}]}])),
 
-    lager:info("checking search 1.0 403s because search won't allow"
-               "connections with security enabled"),
+    lager:info("checking search 1.0 404s because search is removed"),
 
-    ?assertMatch({ok, "403", _, <<"Riak Search 1.0 is deprecated", _/binary>>},
+    ?assertMatch({ok, "404", _, _},
                        ibrowse:send_req(URL ++ "/solr/index/select?q=foo:bar&wt=json", [], get,
```
llelf, line 516: This is a valid yoko URL though. But there's no way that would interfere, right?

martinsumner: The test doesn't start yokozuna, so no that shouldn't cause nay clash.  The equivalent functionality for yoko is tested in yz_security separately.

PR #1317, opened 2019-04-02
Split out intermittent failing part of aae test

martinsumner:
One part of the repl_aae_fullsync test intermittently fails.  Whilst this is worked on, split this part of the test out, so we don't mask other failures being introduced.

I'm not clear how the repl_aae_full_sync_blocked test (the split out part) ever passes.  But intermittently it does.  So more investigation is required.

### DIFF
--- a/groups/bitcask_only
+++ b/groups/bitcask_only
@@ -1,3 +1,2 @@
-verify_bitcask_tombstone2_upgrade
 verify_no_writes_on_read
 kv679_dataloss_fb2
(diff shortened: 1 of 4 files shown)


martinsumner:
I still don't know why this test passes (intermittently).  It looks like I'm achieving < 20% pass rate on the `repl_aae_fullsync_blocked` test.

The scenarios being tested are known to cause problems in production environments (that is to say full-sync fails because one of the vnodes is blocked).  Hence not being sure why this would ever pass.

PR #1324, opened 2019-09-29
Changes to pass some kv_all tests

martinsumner:
riak727 is removed from kv_all upgrade, as the test is a test of an upgrade from 1.4.  It has been stated that for 3.0 upgrades will be supported from 2.9 and 2.9.1 only.

Fix to path in kv679 dataloss to reflect change in riak `make devrel` script which now includes `riak` within the path.

The pb_cipher_suites test has been changed to reflect that protocols before TLS1.2 are no longer supported.  Also, the implementation of honor_cipher_order has switched from patched OTP to mainstream OTP - and as part of this change the default of honor_cipher_order has been switched from false to true.  So now, to honour cipher_order this must be explicitly set to true.  This is a change that *must* be recognised in release notes.  It may have been possible to change the code to implement the previous default behaviour, but having an explicit difference between riak and OTP defaults may cause long term confusion.

https://github.com/erlang/otp/blob/OTP-20.3.8.22/lib/ssl/src/ssl_internal.hrl#L133-L135

https://github.com/erlang/otp/edit/maint/lib/ssl/doc/src/ssl.xml#L1034

The riak_rex test was a test of safe_rpc - that rpc to another node would not throw an exception.  Now, though OTP has changed to stop it from throwing an exception.

https://github.com/erlang/otp/commit/cccf762d93640b707f8e1515604cebcfbe379c04

This means that the safe_rpc try/catch is no longer necessary.  Open question about whether the safe_rpc wrapper should be changed to return the previous error for backward compatibility (although nothin other than the test appears to exist in the code to handle this specific error), or whether the try/catch should just be removed

### DIFF
--- a/tests/pb_cipher_suites.erl
+++ b/tests/pb_cipher_suites.erl
@@ -66,12 +66,12 @@ confirm() ->
     ok = rpc:call(Node, riak_core_console, add_source, [["user", "127.0.0.1/32",
                                                     "password"]]),
 
-    CipherList = "AES128-SHA256:RC4-SHA",
+    CipherList = "AES256-SHA256:RC4-SHA",
     %% set a simple default cipher list, one good one a and one shitty one
     rpc:call(Node, riak_core_security, set_ciphers, [CipherList]),
     rpc:call(Node, application, set_env, [riak_api, honor_cipher_order, true]),
 
-    [AES128, RC4] = ParsedCiphers = [begin
+    [AES256, RC4] = ParsedCiphers = [begin
                 %% this includes the pseudo random function, which apparently
                 %% we don't want
                 SD = ssl_cipher:suite_definition(E),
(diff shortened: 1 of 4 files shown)


tests/pb_cipher_suites.erl
```diff
@@ -66,23 +66,27 @@ confirm() ->
     ok = rpc:call(Node, riak_core_console, add_source, [["user", "127.0.0.1/32",
                                                     "password"]]),
 
-    CipherList = "AES256-SHA256:RC4-SHA",
+    CipherList = "AES128-SHA256:RC4-SHA",
     %% set a simple default cipher list, one good one a and one shitty one
-    rpc:call(Node, riak_core_security, set_ciphers,
-             [CipherList]),
+    rpc:call(Node, riak_core_security, set_ciphers, [CipherList]),
+    rpc:call(Node, application, set_env, [riak_api, honor_cipher_order, true]),
```
ThomasArts, line 72: Here a comment could be welcome. 
If I got it right it means: 
`If true, the Riak node uses its preference for cipher selection, despite what the client indicates.`

ThomasArts (approved):

PR #1326, opened 2019-10-01
Mas i1719 multihead

martinsumner:
Configures a multi-backend node for a test.  Depending on the backend configured in a configuration, sets a bucket to be mapped to that backend (out of the multi-backend).  Then tests the correct choice of HEAD capability.

This confirms for bucket properties (i.e. backend mapping via bucket properties) set via type and directly.


martinsumner:
As it is adding a test (not amending an existing one) I will self + 1 here

PR #1331, opened 2019-10-28
Develop 3.0 ensemble

martinsumner:
These changes are primarily about correcting the test conditions so that the target_n_val is set correctly, and that the number of nodes and n-vals used in the tests will consistently work - without there being a chance that sometimes a minority of nodes might contain a majority of partitions.

There is still an outstanding issue with ensemble_remove_node2 - https://github.com/basho/riak_core/issues/943

There is still a very intermittent failure in `ensemble_sync`.  Very rarely one of the PUTs will timeout https://github.com/basho/riak_test/blob/develop-2.9/tests/ensemble_sync.erl#L109-L110.  It is not the first put.  On one occasion it was associated with a vnode unexpectedly trying to handoff via raw_put and the bang returning a badarg.

Note also - https://github.com/basho/riak/issues/994


ThomasArts (approved):
I am not an esemble expert and cannot really judge the impact, but it seems sound to make these changes. They basically make the tests more stable, don't change the code. Moreover, test behaviour has changed a bit, but not in essence.

PR #1332, opened 2019-10-31
Missed test

martinsumner:
this test was missed out as ensemble_util is not a tets module, and tests in a group don't run if after someting which is not a test module.

the test ensemble_vnode_crash need a correct numbe rof nodes, and also the kill needed to avoid a race condition, and the intercept needed to be loaded on the node being killed.

Strengthened the check to make sure that n-val downed ensemble_backends are seen

### DIFF
--- a/groups/ensemble
+++ b/groups/ensemble
@@ -9,5 +9,4 @@ ensemble_remove_node2
 ensemble_ring_changes
 ensemble_start_without_aae
 ensemble_sync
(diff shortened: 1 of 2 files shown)


martinsumner:
@ThomasArts I was missing a test out on ensemble because of how groups work is you put a non-test in the group ... and the test didn't work.  Fixed now

